PR #3, opened 2019-11-12
Insanity framework

Kmc2000:
Adds a basis for Psy damage as requested by Westhybrid.
Functionally incomplete, up for purposes of review.

Current issues:
Spam text when you see something traumatic
Insanity doesn't really do that much to you, outside of a few hallucinations


config/example/admins.txt
```diff
@@ -5,4 +5,4 @@
 # Ranks can be anything defined in admin_ranks.txt           ~Carn   #
 ######################################################################
 
-# not_a_user - Admin
+Kmc2000 - Admin
```
NanakoAC: You've uploaded your admin config

NanakoAC (changes requested):

code/modules/mob/living/living.dm
```diff
@@ -464,6 +464,11 @@ default behaviour is:
 	reload_fullscreen()
 	return
 
+/mob/living/carbon/human/rejuvenate()//DS13 - Psychosis damage. Overriden as psychosis only occurs on human subtype mobs.
```
NanakoAC, line 467: this human proc doesn't belong in living.dm

baystation12.dme
```diff
@@ -2723,6 +2723,8 @@
 #include "code\unit_tests\zas_tests.dm"
 #include "code\unit_tests\~helpers.dm"
 #include "code\unit_tests\~unit_test_types.dm"
+#include "ds13\code\controllers\subsystems\trauma.dm"
```
NanakoAC, line 2726: these typepaths have gone horribly wrong

code/modules/client/client_procs.dm
```diff
@@ -129,7 +129,12 @@
 			log_admin("[ckey] tried to join and was turned away due to the server being full (player_limit=[config.player_limit])")
 			qdel(src)
 			return
-
+	//DS13 - Give locally logged in users host status
+	var/localhost_addresses = list("127.0.0.1", "::1")
+	if(isnull(address) || (address in localhost_addresses))
+		var/rights = admin_ranks["Host"]
+		var/datum/admins/D = new /datum/admins("Host", rights, ckey)
+		D.associate(src)
```
NanakoAC, line 137: does this localhost thing work now ?I recall you had a lot of problems with it

code/modules/mob/living/carbon/human/species/necromorph/mob.dm
```diff
@@ -5,6 +5,7 @@
 `	I am well aware this is not how human mobs and species are supposed to be used
 */
 /mob/living/carbon/human/necromorph
+	psychosis_immune = TRUE //Because..y'know.
```
NanakoAC, line 8: This should be set via species, not directly on the mob.
Usage of necromorph mob subtypes is temporary, everything works through species code

config/example/admins.txt
```diff
@@ -5,4 +5,4 @@
 # Ranks can be anything defined in admin_ranks.txt           ~Carn   #
 ######################################################################
 
-# not_a_user - Admin
+# not_a_user - Admin
```
NanakoAC, line 8: how did this get uploaded

ds13/code/modules/mob/living/psychosis.dm
```diff
@@ -0,0 +1,182 @@
...
+
+
+// Psychosis side effects:
+// ========
+
+
+/datum/medical_effect/irrational_laughter
+	name = "Irrational laughter"
+	triggers = list() //Triggered only by psychosis
+	cures = list(/datum/reagent/synaptizine) //Cured with brain meds
```
NanakoAC, line 22: a minor thing, but this should probably be cured with sedatives

ds13/code/modules/mob/living/psychosis.dm
```diff
@@ -0,0 +1,182 @@
...
+/**
+*
+*
+* A proc which handles fluff FX for people going mad.
+* @Param stage : How severe the psychosis is. This is a scaling value based on the mob's current psychosis damage (which is increased via witnessing traumatic events, or being touched by necros).
+*
+*/
+
+/mob/living/carbon/human/proc/set_psychosis(var/stage, var/override)
+	if(!can_set_psychosis && !override) //Avoids patients being permanently debilitated by psychosis such that they can't move. EG: if youre on 0 sanity, you have a small window to get help before the effects manifest again.
```
NanakoAC, line 53: what's with this can_set_psychosis var? It seems a bit heavy handed. Can you explain how it works ?

ds13/code/modules/mob/living/psychosis.dm
```diff
@@ -0,0 +1,182 @@
...
+* A series of procs to handle psychosis damage on mobs.
+*
+*/
+
+/mob/living/carbon/human
+	var/sanity = 100 //Flat value to represent how ""sane"" someone is. Losing sanity is not recommended.
+	var/maxSanity = 100
+	var/last_sanity = 100
+	var/can_set_psychosis = TRUE //This variable means that mobs can't instant descend into total insanity just by getting sneezed on by a necromorph.
+	var/psychosis_immune = FALSE //SET THIS TO TRUE FOR NECROMORPHS IF YOU MAKE THEM A HUMAN SUBTYPE, OR THEY WILL GO INSANE FROM MARKER STUFF
```
NanakoAC, line 94: Rather than a flat value, i recommend making this a multiplier, so partial resistance can be a thing too

ds13/code/modules/mob/living/psychosis.dm
```diff
@@ -0,0 +1,182 @@
...
+	var/sanity_cooldown = FALSE
```
NanakoAC, line 95: using extensions for this may be better, adding too many vars to mobs isn't a good idea

ds13/code/modules/mob/living/psychosis.dm
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+*
+*
+* A proc which handles fluff FX for people going mad.
+* @Param stage : How severe the psychosis is. This is a scaling value based on the mob's current psychosis damage (which is increased via witnessing traumatic events, or being touched by necros).
+*
+*/
+
+/mob/living/carbon/human/proc/set_psychosis(var/stage, var/override)
```
NanakoAC, line 52: I'm not clear on the purpose of this function at all, it doesn't seem to use the granular psychosis loss which is healed and set elsewhere, and the effects should probably not be embedded right here

ds13/code/modules/mob/living/psychosis.dm
```diff
@@ -0,0 +1,182 @@
...
+	var/name = "Traumatic sight handler"
+	var/atom/source = null
+	var/trauma_power = 1.50 //How traumatic is this sight? Dealt to them per traumatic object, per 10 seconds. A room full of necros would seriously fuck you up. A "STRONG" trauma would have a value of 5, and would give you just over a minute to cover your eyes before it started to affect you. Better bring that blindfold with you!
+
+/datum/component/traumatic_sight/New()
+	. = ..()
+	START_PROCESSING(SStrauma, src)
+
+/datum/component/traumatic_sight/Process() //This isn't terribly efficient, but it's also not the worst performance wise.
+	for(var/mob/living/carbon/human/H in viewers(source, null))
```
NanakoAC, line 166: did you use step triggers like i advised? Having these things polling regularly is not good.
I realise its a slow polling, but that also means things are very unresponsive.

NanakoAC (changes requested):

ds13/code/modules/mob/living/psychosis.dm
```diff
@@ -0,0 +1,182 @@
...
+*
+* A proc which handles fluff FX for people going mad.
+* @Param stage : How severe the psychosis is. This is a scaling value based on the mob's current psychosis damage (which is increased via witnessing traumatic events, or being touched by necros).
+*
+*/
+
+/mob/living/carbon/human/proc/set_psychosis(var/stage, var/override)
+	if(!can_set_psychosis && !override) //Avoids patients being permanently debilitated by psychosis such that they can't move. EG: if youre on 0 sanity, you have a small window to get help before the effects manifest again.
+		return
+	switch(stage)
```
NanakoAC, line 55: this switch statement is definitely not a good design as is

rather than strictly giving irrational laughter at tier 1, i'd recommend organising the effects into a few tier lists, and picking randomly from those lists

even if there's only one thing for now, the system should be expandable to many possible effects later

PR #4, opened 2019-11-16
Necromorphs Phase 1:

NanakoAC:
Overview of PR contents
	Necromorph Framework
	Slasher
	Door forcing, and various door balances
	Charge attack
	Hotkey verb system
	Debug settings
	
1. Adds a basic framework for necromorphs to use going forward. They are based off the human type, and implemented as a species (and many subspecies).
This does all the fundamental work to implement them. Making them not breathe, immune to pressure, vulnerable to fire, etc. They have all the same organs as a human, but the heart and lungs are special dead versions which don't need to function. They bleed brown, but don't need blood to live and will keep going even if its all drained
	There is some code on the mob which is used for their icon updates, and thus they will not work properly if the correct mob types aren't spawned. This functionality is temporary, and once we have the sprites for it, they will use the normal human method of being built out of a sprite-per-limb, and thus able to be dismembered. For now, dismemberment does not function properly.
	
	
2. Slasher
	Adds the first necromorph, the Slasher, using the dessicated slasher sprites.
	The slasher has two huge blades instead of arms, and these are properly represented ingame. Its basic attack is a powerful cleave which can chop people up fairly effectively. The damage may be a bit high, currently set to 15 at west's direction, it may need rebalancing in in future. Currently it can decapitate an unprotected human in eight hits, (and cause probably-fatal bleeding in a bit less)
	
	The slasher's blade arms are not hands, and so it can't pickup objects. Its health is currently set at 80 (a normal human has 120)
	
3. Door forcing
	How airlocks break is significantly reworked. Now an airlock which runs out of health will break -open-, meaning that if you keep hitting an airlock hard enough it will eventually give way. Airlocks have a resistance which subtracts a flat quantity from the damage of incoming hits, meaning weaker blows will never be able to harm it.  The incoming damage is also reduced (before resistance) by being bolted or welded, thus these can make it last longer
	
	Secondly, certain unarmed attacks (which includes the slasher's blades) can now force open an airlock when used in grab intent. This is generally much faster than breaking it with repeated hits. Various locks and welds increase the difficulty and time required to do this.
	
	Thirdly, as requested, support for door reinforcement has been partially implemented. We have no sprites yet, so this needs a bit more work before it can actually be doable ingame, but the door forcing mechanics account for it and the internal plumbing is there.
	
	
4. Charge Attack
	Adds a well implemented charging attack which can be added to any mob or human species. 
	Charge uses an extension that takes a variety of parameteres, power, speed, time, homing, etc. It moves the user towards the target at a rapid pace, triggering events with anything they bump into along the way. These use new procs, charge_impact (on the charger) and charge_act (on the reciever) which can be implemented and overrridden by any atom to do some custom behaviour. 
	
	The default behaviour on objects routes charge_act to ex_act which serves pretty well for most purposes. allowing charging mobs to break through lightweight obstacles, like plastic tables. The power var controls the strength of this, and a high power charge could break through anything. On mobs, charge_act is instead translated to damage as a melee hit.
	
	The slasher gets a custom charge impact, instead of the above, it gets two free melee attacks on the target(one per remaining blade arm, so dismembering helps weaken it)
	
	
5. Hotkey verbs
	Implements a lightweight and excessively flexible system for allowing verbs to be easily mapped to modifier keys. Currently implemented for human species, but it wouldn't be hard to add to any other controllable mob. Allows any verb to be mapped to altclick, shiftclick, ctrlclick, middleclick, ctrl+altclick, and ctrl+shiftclick
	
	The slasher is given charge mapped to altclick. It only triggers on atoms that are at least two tiles away, so altclick can still be used to examine the contents of adjacent tiles
	
	
6. Debug Settings
	Ports some debug settings i wrote for another server. This adds a new config file, debug.txt. It contains several options designed to make starting up a test server much faster and require fewer clicks. All the options in it should be disabled for live production code. Includes:
		Autostart: Starts the round ASAP
		Auto-observe: Automatically spawns you as an observer
		Auto BST (Requires auto observe) automatically spawns you as a bluespace tech
		No overmap: Disables overmap and exoplanet generation, significantly reducing server startup time
		
	With all of these options enabled, you can hit Build and Run in dream maker with one click, and then come back 2 minutes later to find yourself standing in the cryo room, ready to do things. Having personally done this process over a thousand times, making it less tedious is a big deal


code/_helpers/matrices.dm
```diff
@@ -19,8 +19,48 @@
...
+		if(!M.client)
+			return
+
+		var/px_y = rand(0, strength*world.icon_size) * pick(-1, 1)
+		var/px_x = rand(0, strength*world.icon_size) * pick(-1, 1)
+		var/vector2/init_px = new /vector2(M.client.pixel_x, M.client.pixel_y)
+		animate(M.client, pixel_x=init_px.x + px_x, pixel_y=init_px.y + px_y, time=1)
+		animate(pixel_x=init_px.x, pixel_y=init_px.y, time=duration, easing=ELASTIC_EASING)
+
+	/*(
```
Kmc2000, line 44: Remove this?

code/__defines/_tick.dm
```diff
@@ -19,3 +19,10 @@
 //for when you need a reliable time number that doesn't depend on byond time.
 #define REALTIMEOFDAY (world.timeofday + (MIDNIGHT_ROLLOVER * MIDNIGHT_ROLLOVER_CHECK))
 #define MIDNIGHT_ROLLOVER_CHECK ( GLOB.rollovercheck_last_timeofday != world.timeofday ? update_midnight_rollover() : GLOB.midnight_rollovers )
+
+
+//Takes a speed in metres per second, and outputs delay in deciseconds between each step to achieve that
```
NanakoAC, line 24: i made two copies of this define, need to delete one

code/_helpers/matrices.dm
NanakoAC, line 44: oops forgot to delete old camera shake code

I've revamped camera shaking to use animate, its much better

code/game/gamemodes/gameticker.dm
```diff
@@ -76,6 +84,22 @@ var/global/datum/controller/gameticker/ticker
 	while (!setup())
 
 
+/datum/controller/gameticker/proc/start_now(mob/user)
+	//if(!(current_state == RUNLEVEL_LOBBY))
```
NanakoAC, line 88: commented code to remove here

code/modules/mob/living/abilities/charge.dm
```diff
@@ -0,0 +1,388 @@
...
+		Target: Must be an atom with a non-null location on the same zlevel. Required
+		Speed: How fast to move this mob towards the target, in metres (tiles) per second. Required
+		Time: Maximum time in deciseconds we will spend attempting to reach the target. If this much time elapses, the charge is aborted
+			A time is needed, default is 2 seconds
+		Range: Maximum number of tiles we will cross while chasing the target. If we meet this range the charge is aborted
+			Optional, the default value of 0 disables range limits
+		Homing: If true, we will keep moving towards the target even when they move.
+			If false, we will aim at the spot they were standing when we started, and miss them if they aren't in it
+			Optional, Default True
+		Inertia: Only used if homing is false.
```
NanakoAC, line 22: need to add note that inertia is currently unimplemented

code/modules/mob/living/carbon/human/species/necromorph/necromorph.dm
```diff
@@ -0,0 +1,105 @@
...
+	BP_HEART =    /obj/item/organ/internal/heart/undead,
+	BP_LUNGS =    /obj/item/organ/internal/lungs/dead,
+	BP_LIVER =    /obj/item/organ/internal/liver,
+	BP_KIDNEYS =  /obj/item/organ/internal/kidneys,
+	BP_BRAIN =    /obj/item/organ/internal/brain,
+	BP_APPENDIX = /obj/item/organ/internal/appendix,
+	BP_EYES =     /obj/item/organ/internal/eyes
+	)
+
+	/* //TODO: Uncomment this once limbs are sorted
```
NanakoAC, line 86: remove this, i ended up using limbs anyway

code/modules/mob/living/carbon/human/species/necromorph/slasher.dm
```diff
@@ -0,0 +1,101 @@
...
+
+/atom/movable/proc/slasher_charge(var/atom/A)
+	set name = "Charge"
+	set category = "Abilities"
+
+
+	.= charge_attack(A, _delay = 1.5 SECONDS)
+	if (.)
+		face_atom(A)
+		//Do some audio cues here
```
NanakoAC, line 74: Need audio for this

code/modules/mob/living/carbon/human/unarmed_attack.dm
```diff
@@ -112,6 +129,91 @@ var/global/list/sparring_attack_cache = list()
...
+		var/time = target.get_force_time()
+
+		//Now, if we have more power than we need, we get a discount on that time
+		if (airlock_force_power > difficulty)
+			time /= 1 + (airlock_force_power - difficulty)
+
+		//And we divide by our speed too
+		time /= get_force_speed(user, target)
+
+		//Ok lets start
```
NanakoAC, line 192: Need metal straining sounds for this

code/game/machinery/doors/airlock.dm
```diff
@@ -1007,121 +1007,121 @@ About the new airlock wires panel:
...
-				src.welded = 1
+		if (!repairing && (stat & BROKEN) && src.locked) //bolted and broken
+			if (!cut_bolts(C,user))
+				..()
+			return
+
+		if(!repairing && isWelder(C) && !( src.operating > 0 ) && src.density)
+			user.visible_message("You start [welded?"unwelding":"welding"] the [src]")
+			if(C.use_tool(user, src, WORKTIME_VERY_SLOW, QUALITY_WELDING, FAILCHANCE_NORMAL))
+				if(!src.welded)
```
NanakoAC, line 1047: We need sprites for reinforced doors here

code/_onclick/click.dm
```diff
@@ -287,7 +286,17 @@
...
 	Only used for swapping hands
 */
-/mob/proc/MiddleClickOn(var/atom/A)
+/mob/proc/MiddleClickOn(var/atom/A, var/params)
+	var/datum/stack/click_handlers
+
+	click_handlers = src.GetClickHandlers()
+
+
+	while (click_handlers.Num())
```
Kmc2000, line 295: While i'm not at all familiar with click handlers on bay, having a while loop every time someone middle clicks seems odd..
Could you convert this to signals / bay's equivalent instead?

code/controllers/subsystem.dm
```diff
@@ -163,8 +163,21 @@
...
-	to_chat(world, "<span class='boldannounce'>[msg]</span>")
-	log_world(msg)
+
+	//The formatting of outputs depends on the time taken
+	//Longer than one second, its in red
+	//The formatting of outputs depends on the time taken
+	//Longer than one second, its in red
+	if (time > 1)
+		//log_world(msg)
+		report_progress("<span class='boldannounce'>[msg]</span>")
```
Kmc2000, line 173: Will this not spam up the discord if this is run on live? Or am I misreading report_progress()

code/datums/autolathe/autolathe_datums.dm
```diff
@@ -38,17 +38,21 @@ other types of metals and chemistry for reagents).
 
 //These procs are used in subtypes for assigning names and descriptions dynamically
 /datum/design/proc/AssembleDesignInfo()
+
```
Kmc2000, line 41: Possible unstated change?

code/game/gamemodes/gameticker.dm
```diff
@@ -76,6 +84,22 @@ var/global/datum/controller/gameticker/ticker
...
+	initialization_stage |= INITIALIZATION_NOW
+	bypass_gamemode_vote = TRUE
+	vote.reset()
+	//if(istype(SSvote.active_vote, /datum/vote/gamemode))
+	//	SSvote.cancel_vote(user)
+
+	//Master.SetRunLevel(RUNLEVEL_SETUP)
+
+
+	return 1
```
Kmc2000, line 101:
```suggestion
   return TRUE
```

code/game/gamemodes/gameticker.dm
```diff
@@ -151,6 +175,17 @@ var/global/datum/controller/gameticker/ticker
 
 	callHook("roundstart")
 
+	//Here we will trigger the auto-observe and auto bst debug things
```
Kmc2000, line 178: This seems hacky. What is this "auto observe" and "auto BST" stuff?

code/game/machinery/doors/airlock.dm
```diff
@@ -1414,4 +1414,50 @@ About the new airlock wires panel:
...
+
+
+/obj/machinery/door/airlock/get_force_difficulty()
+	.=..()
+	if(!arePowerSystemsOn() || isWireCut(AIRLOCK_WIRE_OPEN_DOOR))
+		.-=(force_resist* 0.3)//Depowered door is easier to force, it has a third of its base force resistance
+
+	//Bolts help
+	if (locked)
+		. += 0.5
```
Kmc2000, line 1427: That's neat, I'll take a note that you can do that

code/game/machinery/doors/door.dm
```diff
@@ -522,3 +571,12 @@
 			dirs |= direction
 	connections = dirs
 
+
+//For forcing
+/obj/machinery/door/proc/get_force_difficulty()
+	.=force_resist
```
Kmc2000, line 577:
```suggestion
. = force_resist
```

code/game/machinery/doors/door.dm
```diff
@@ -522,3 +571,12 @@
...
+
+
+
+/obj/machinery/door/proc/get_force_time()
+	.= force_time * get_force_difficulty()
```
Kmc2000, line 582:
```suggestion
. = force_time * get_force_difficulty()
```

code/game/world.dm
```diff
@@ -521,6 +521,7 @@ var/world_topic_spam_protect_time = world.timeofday
 /proc/load_configuration()
 	config = new /datum/configuration()
 	config.load("config/config.txt")
+	config.load("config/debug.txt")
```
Kmc2000, line 524: Intentional?

code/modules/admin/bluespacetech.dm
```diff
@@ -57,10 +59,9 @@
 	bst.update_inv_wear_id()
 	bst.regenerate_icons()
 
-	//BSTs have all languages
-	bst.add_language(LANGUAGE_GALCOM)
-	bst.add_language(LANGUAGE_GUTTER)
-	bst.add_language(LANGUAGE_SIGN)
+	//TODO:
```
Kmc2000, line 62: Commented code

code/modules/admin/bluespacetech.dm
```diff
@@ -266,7 +265,8 @@
 	name = "bluespace technician's shoes"
 	desc = "A pair of black shoes with extra grip. The letters 'BST' are stamped on the side."
 	icon_state = "black"
-	item_flags = ITEM_FLAG_NOSLIP
+	//TODO: Enable noslip
```
Kmc2000, line 268: Commented code

code/modules/admin/bluespacetech.dm
```diff
@@ -310,12 +310,15 @@
 	return fall_override ? FALSE : ..()
 
 
+//These are just wrappers on the standard move up/down verbs, duplicating them into the BST menu for easy clicking
 /mob/living/carbon/human/bst/verb/moveup()
 	set name = "Phase Upwards"
 	set category = "BST"
-	zMove(UP)
+	up()
+	//zMove(UP)
```
Kmc2000, line 318: Commented code

code/modules/mob/living/abilities/charge.dm
```diff
@@ -0,0 +1,388 @@
...
+#define CHARGE_DAMAGE_DIST	2	//Extra damage dealt by charge per tile travelled while charging
+
+//States of a charge
+#define CHARGE_STATE_WINDUP		"windup"	//We're preparing to charge, maybe screaming a bit
+#define CHARGE_STATE_CHARGING	"charging"	//We're running towards a target
+#define CHARGE_STATE_COOLDOWN	"cooldown"	//We're cooling after a charge
+/*
+	The charge extension is attached to a mob and acts as a puppeteer to move it
+*/
+/datum/extension/charge
```
Kmc2000, line 44: Note to self:
Example of extensions

code/modules/mob/living/abilities/charge.dm
```diff
@@ -0,0 +1,388 @@
...
+
+	if (homing)
+		//If homing, we aim straight for the thing
+		move_target = target
+	else if (!inertia)
+		//If no homing or inertia, we aim for the tile the thing is on
+		move_target = get_turf(target)
+	else
+		//If no homing but yes inertia, we project a straight path out as far as we can
+		/*var/turf/T1 = get_turf(holder)
```
Kmc2000, line 112: Commented code

code/modules/mob/living/abilities/charge.dm
```diff
@@ -0,0 +1,388 @@
...
+//------------------------
+/atom/movable/proc/charge_verb(var/atom/A)
+	set name = "Charge"
+	set category = "Abilities"
+
+
+	return charge_attack(A)
+
+
+/atom/movable/proc/can_charge(var/atom/target, var/error_messages = TRUE)
```
Kmc2000, line 341: Would you not want this on the extension itself? With a verb added by the extension when they're granted "charge"? We can probably assume that they're not going to be able to charge without said extension.

code/modules/mob/living/carbon/human/human_attackhand.dm
```diff
@@ -114,11 +114,26 @@
...
 		if(I_GRAB)
+			if (!can_grasp_with_selected())
+				to_chat(H, "<span class='warning'>You can't use your hand.</span>")
+				return
 			return H.species.attempt_grab(H, src)
 
 		if(I_HURT)
 
-			if(!istype(H))
+			//This whole thing below here is a mess.
```
Kmc2000, line 124: Baycode summed up nicely 😎

code/modules/mob/living/carbon/human/species/necromorph/mob.dm
```diff
@@ -0,0 +1,35 @@
+/*
+	Code for the necromorph mob.
+	Most of this is a temporary hack because we don't have proper icons for parts.
+
+`	I am well aware this is not how human mobs and species are supposed to be used
+*/
+/mob/living/carbon/human/necromorph
```
Kmc2000, line 7: I'm really unsure about having necros be humans...
Do they need to be able to pick up things? Would simplemobs not make this much easier?

code/modules/mob/living/carbon/human/species/species_attack.dm
```diff
@@ -5,11 +5,9 @@
 	sharp = 1
 	edge = 1
 
-/datum/unarmed_attack/diona
-	attack_verb = list("lashed", "bludgeoned")
-	attack_noun = list("tendril")
-	eye_attack_text = "a tendril"
-	eye_attack_text_victim = "a tendril"
+
```
Kmc2000, line 8: In future, I'd recommend something like:
//DS13 - Moved to our modular folder.
Or just a straight up override, deletion like this shouldn't be necessary

code/modules/mob/mob_helpers.dm
```diff
@@ -314,28 +314,7 @@ It's fairly easy to fix if dealing with single letters but not so much with comp
 	return sanitize(t)
 
 
-/proc/shake_camera(mob/M, duration, strength=1)
```
Kmc2000, line 317: Again, a comment showing to where this has moved might be useful

Kmc2000 (commented):
Few issues, but the bulk of it SEEMS to be good!
This is by no means exhaustive, though

code/controllers/subsystem.dm
NanakoAC: the total number of logs is less than before

code/datums/autolathe/autolathe_datums.dm
NanakoAC: ah this was an attempt to fix a runtime error, but reverted, its got to be done later

code/game/gamemodes/gameticker.dm
NanakoAC: done

code/game/gamemodes/gameticker.dm
NanakoAC: Debug settings, mentioned in the description

code/game/world.dm
NanakoAC: yes

code/modules/admin/bluespacetech.dm
NanakoAC: a todo isn't code, its a reminder for future features. BSTs are for debugging so this is low priority

code/modules/mob/living/abilities/charge.dm
NanakoAC: No, its overrideable so that various mob types can do their own checks. There's a human version just a few lines down from that

code/modules/mob/living/carbon/human/species/necromorph/mob.dm
NanakoAC: This is settled, we spent three days mulling it over collectively. Human subtype is needed in the future for dismemberment

code/modules/mob/living/carbon/human/species/species_attack.dm
NanakoAC: i was just tidying a few things up here

code/modules/mob/mob_helpers.dm
NanakoAC: DS13 no longer inherits from bay, its not really intended to be used as a base by other repos

This proc was silly to exist in mob helpers to begin with, since
1. Its not a mob proc, its global
2. It works on the client, not the mob

Kmc2000: This is true, but i'm more talking about tracking changes internally and it's good practice anyway.

PR #6, opened 2019-11-17
Audio and slasher fix

NanakoAC:
Fixes a compile error and a critical bug


NanakoAC:
This is a minor and uncontroversial patch which fixes two critical bugs

PR #10, opened 2019-11-30
Necromorphs Phase 2: Brute, vision improvements, species code

NanakoAC:
This PR adds the necromorph brute, far more complicated than the slasher, and thus accompanied by a vast quantity of fixes and infrastructure code to make its various features work.

First up, the brute itself is a huge creature. It still takes up only a single tile for its collision box, but it is largely an immovable object. Ordinary humans cannot push it aside or pull it, it blocks movement even when downed, and it blocks vision, making you unable to see what hides behind it.

The brute has permanant tunnel vision, causing it to have moderately reduced view distance to its sides, and massively reduced rear vision. Also includes a smooth saccade effect when switching view angles. This took a looot of effort to get working, varying vision radiuses are a native byond feature, but baycode had a lot of bugs which broke it, and i had to fix all of those. This particular feature may be worth sending upstream to bay. This also means we now have full support for both varying view radiuses, and offsetting the view forward or back. This could be used in a lot of other ways, like ironsights aiming or vision enhancing helmets allowing people to see farther. Its a feature with tons of future potential
Brute Vision

Its primary attack is a powerful knockback punch, dealing a basic hit of 25 damage, but also violently hurling the victim away. Mobs thrown in this manner will bounce hard off walls and similar dense objects, suffering additional damage. The sound effects are pretty satisfying
Brute Punch

The brute has a unique secondary attack, Slam. It rears up and back, before bringing both arms down, dealing a shockwave of damage in a 3-wide, 2-long arc infront of itself. Mobs hit by this attack take heavy damage and get knocked down, the damage is doubled if they were already down when hit. It can also smash through many obstacles in the way. Tables and windows will crumble to dust, repeated hits may eventually break through airlocks and even walls. Slam is slow and easy for a mobile human to dodge, but devastating if you can land it.
Slam can be activated quickly using Ctrl+Alt+Click
Brute Slam

The slasher's Charge attack also makes a return. The brute's variant is a bit slower, but hits harder, and has inertia activated (I fixed this!). The inertia flag will cause it to keep charging until something gets in the way, which can be both a blessing and a curse. If used well, it can cover long swathes of open ground relatively quickly, but it is also quite easy for a human to dodge out of the path, and perhaps shoot the brute in the back as it surges past. Charge is a double edged sword, and a useful tool in its arsenal.
Charge can be activated quickly using Alt+Click
Brute Charge

The brute's front and side are covered in heavy organic bone plating, which has a 95% chance to intercept any incoming hit. Attacks caught in this way have their damage reduced by a flat value (front has stronger armor than sides). Melee attacks that are completely blocked in this way will rattle the attacker, while blocked projectiles will richochet off the brute and potentially hit other things around. Getting directional armor functioning required a fair bit of effort adding in species-level intercepts on damage functions, but i have done it properly, with no mob code which would break the species design spec.
Brute Armor

The brute's rear is completely unprotected, and if it takes a hit which isn't caught by the armor, it will reflexively curl up for a few seconds. The Curl ability can also be used manually at any time, and it renders the brute unable to move or turn and limits its vision farther, but increases the strength of the armor by 75%, and increases the front+side coverage to 100%. Curl also comes with a neat sprite animation, i'm good at these
Curl can be activated quickly using Ctrl+Shift+LMB
Brute Curl

Lastly, this PR also includes lots of tiny tweaks to balance necromorphs a bit, various fixes and things that i've forgotten. will try to add to this description


code/_onclick/hud/human.dm
```diff
@@ -150,6 +150,8 @@
 		using.alpha = ui_alpha
 		src.hotkeybuttons += using
 
+
```
Kmc2000, line 153: Whitespace

code/game/objects/structures.dm
```diff
@@ -113,4 +113,26 @@
 			other_dirs += get_dir(src, T)
 
 	connections = dirs_to_corner_states(dirs)
-	other_connections = dirs_to_corner_states(other_dirs)
\ No newline at end of file
+	other_connections = dirs_to_corner_states(other_dirs)
+
+
+// When destroyed by explosions, properly handle contents.
+obj/structure/ex_act(severity, var/atom/epicentre)
```
Kmc2000:
```suggestion
/obj/structure/ex_act(severity, var/atom/epicentre)
```

code/game/objects/structures/transit_tubes.dm
```diff
@@ -56,26 +56,7 @@
 
 
 
-// When destroyed by explosions, properly handle contents.
```
Kmc2000, line 59: Intentional?

code/game/turfs/simulated/floor_acts.dm
```diff
@@ -17,6 +15,13 @@
 					else
 						src.break_tile()
 					src.hotspot_expose(1000,CELL_VOLUME)
+		if(2.0)
+			if(prob(33)) new /obj/item/stack/material/steel(src)
```
Kmc2000:
```suggestion
			if(prob(33)) 
				new /obj/item/stack/material/steel(src)
```

code/modules/client/client_helpers.dm
```diff
@@ -12,3 +12,37 @@
...
+
+
+/client/proc/set_view_offset(var/direction, var/magnitude)
+	var/vector2/offset = (Vector2.FromDir(direction))*magnitude
+	if (pixel_x != offset.x || pixel_y != offset.y) //If the values already match the target, don't interrupt the animation by repeating it
+		animate(src, pixel_x = offset.x, pixel_y = offset.y, time = 5, easing = SINE_EASING)
+
+
+/client/proc/setup_click_catcher()
+	if(!void)
```
Kmc2000, line 34: Void?

Kmc2000 (approved):
Looks good to me. I can't see anything majorly wrong with it so it should be good for testing.

code/game/objects/structures/transit_tubes.dm
NanakoAC: yep, i moved it to a better file
this is a generic structure proc and it definitely shouldn't be in a specific file like this

code/modules/client/client_helpers.dm
NanakoAC: i have no clue, i moved this function from elsewhere

PR #12, opened 2019-12-02
Necromorphs Phase 3: Twitcher

NanakoAC:
This PR adds the necromorph Twitcher, with animated sprites from ketrai.
The twitcher is a variant of slasher, with a stasis module embedded, which makes it do wierd phased time stuff. I've taken a few liberties to reflect this in several ways. Overall, it is somewhat of an elite unit, designed to fill a fast attack and skirmisher role. Something of a mid tier necromorph

First of all, the twitcher has higher base movespeed, and a lower attack delay, compared to slasher. It still has the same health and damage though.
Twitcher also has a view offset allowing it to see a few extra tiles ahead, useful for longrange charges.

Twitcher comes with some unique idle animations which trigger periodically every few seconds, and also every time it binks
twitching

The twitcher has a unique charge attack which, aside from being a bit faster than normal, also kicks off with a short blink, almost faster than the eye can see. Sure to make people panic
twitcher charge

Secondly, the twitcher has a chance (4%) to blink one tile in a random direction with each step it takes. This can make its movements unpredictable.  I was unable to get a clip of this happening in isolation, but I found, to pleasant surprise, that it can also trigger while charging, creating a horrifying event like this:
blink charge

try shooting that!

Lastly, the twitcher also has a defensive blink. This triggers automatically when it takes damage from anything, displacing a tile, and nullifying most (but not all) of the damage from that hit. This effect has a cooldown of 5 seconds, so big slow weapons will not do well against them, but rapid fire attacks will tear them up.  Pulse rifle would be a good choice if you can aim :D
defensive blink


code/_helpers/matrices.dm
```diff
@@ -137,6 +137,50 @@
 
 	return list(R + x,R,R, G,G + x,G, B,B,B + x)
 
+
+//This proc moves mover to target instantly, but plays an animation to make it appear like they're sliding there smoothly.
```
Kmc2000, line 141: I'm not 100% on this implementation, but it'll do. A possible exploit for this might be right clicking your tile constantly to check for movers jumping to you, then running away accordingly?

code/_helpers/turfs.dm
```diff
@@ -20,6 +20,7 @@
 			return 0
 	return 1
 
+
```
Kmc2000: Random whitespace

Kmc2000 (approved):

code/_helpers/matrices.dm
NanakoAC: ive used this before to considerable success, i don't think there'll be issues with it.

In all the cases it's used here, the movement happens within 0.2 seconds anyways. even if someone were theoretically able to do that, they'd never react in time

PR #13, opened 2019-12-06
Brute and Slasher audio

NanakoAC:
This PR first of all adds a swathe of minor new features to accomodate species level audio, and datum-level audio cooldowns in a neat package. It includes significant refactoring of how footsteps work, and folds a few old species vars (death sounds, speech sounds) into the new system.  This system allows audio files for a variety of categories. Pain, death, shouts, long shouts, footsteps, speech, and melee attacks. These are all factored in at appropriate places

Also includes a small bugfixes for necromorphs being given silly random names (like Wahneh Blahblah)

Secondly, it adds TONS of audio files for the brute and slasher. Loud screams to telegraph special attacks, pain and death sounds, screams of rage with attacks.
The brute has big stompy footstep sounds too, however the slasher does not. No footstep sounds for slasher yet.
Slasher has occasional speech audio, though brute doesn't.

Lastly, as an additional bonus, this adds another new special ability, given to these two necros (and most likely, to all of them eventually. ) Shout.

It makes use of the new audio assets to play a loud shouting sound, as well as causing some screenshake for everyone nearby, and briefly stunning the user. No other effects, it's basically a taunt and mostly for roleplay value.

In future, it may be good to add psychosis damage to it too


code/game/objects/structures.dm
```diff
@@ -11,11 +11,21 @@
...
 	var/list/footstep_sounds	//footstep sounds when stepped on
+	var/step_priority = 1	//Priority of the sound attached to this
 
 /obj/structure/proc/get_footstep_sound()
 	if(LAZYLEN(footstep_sounds)) return pick(footstep_sounds)
 
+/obj/structure/New()
+	.=..()
+	if(LAZYLEN(footstep_sounds) && istype(loc, /turf/simulated/floor))
+		var/turf/simulated/floor/T = loc
```
Kmc2000:
```suggestion
		var/turf/simulated/floor/T = get_turf(src)
```

code/game/objects/structures.dm
```diff
@@ -11,11 +11,21 @@
...
+		T.step_structures |= src
+
 /obj/structure/Destroy()
+	if(LAZYLEN(footstep_sounds) && istype(loc, /turf/simulated/floor))
+		var/turf/simulated/floor/T = loc
```
Kmc2000:
```suggestion
		var/turf/simulated/floor/T = get_turf(src)
```

code/game/turfs/simulated/footsteps.dm
```diff
@@ -108,37 +110,48 @@
...
-		if(footsound) break
-	if(!footsound)
+	var/best_priority = 0
+
+	if (species.step_priority > best_priority)
+
+		footsound = get_species_audio(SOUND_FOOTSTEP)
+		best_priority = species.step_priority
+
+	for(var/obj/structure/S as anything in T.step_structures)
```
Kmc2000, line 137: Not the biggest fan of this if I'm honest. A for loop every time you walk sounds expensive. Could you not use something like a step trigger/have a component on the floor which generates the priority step sound at gen / destruction?

code/modules/mob/living/abilities/shout.dm
```diff
@@ -0,0 +1,15 @@
+//Simple ability that makes a loud screaming noise, causes screenshake in everyone nearby.
+/mob/proc/shout()
+	set name = "Shout"
+	set category = "Abilities"
+	if (check_audio_cooldown(SOUND_SHOUT))
+		if (play_species_audio(src, SOUND_SHOUT, 100, 1, 2))
+			src.Stun(2)
+			src.shake_animation(40)
+			set_audio_cooldown(SOUND_SHOUT, 8 SECONDS)
+			for (var/mob/M in range(7, src))
```
Kmc2000, line 10:
```suggestion
			for (var/mob/M in hearers(7, src))
```

code/modules/mob/living/carbon/human/human_helpers.dm
```diff
@@ -300,3 +300,5 @@
 
 	UNSETEMPTY(cloaking_sources)
 	return !cloaking_sources // If cloaking_sources wasn't initially null but is now, we've uncloaked
+
```
Kmc2000, line 303: Whitespace change

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -154,3 +154,10 @@
 				visible_message("<span class='notice'>\The [src] is sweating heavily!</span>", "<span class='notice'>You are sweating heavily!</span>")
 			if(2)
 				visible_message("<span class='notice'>\The [src] looks out of breath!</span>", "<span class='notice'>You are out of breath!</span>")
+
+
+
+
+/mob/living/carbon/human/set_move_intent(var/decl/move_intent/M)
+	.=..()
```
Kmc2000:
```suggestion
	. = ..()
```
For neatness

Kmc2000: (If this is even necessary at all, as you defined it on /mob anyway)

Kmc2000 (approved):
Overall seems like a nice system. I won't comment on the audio files, though, but I'll assume they're OK as WestHybrid provided them.

code/game/turfs/simulated/footsteps.dm
NanakoAC: Have a look at how the code was before. I've massively optimised this loop by creating a new list solely to hold things that affect footsteps. The "as anything" is also a farther optimisation i learned from consulting the byond coding gurus. 

This is massively less expensive than it used to be, and about as good as it can get

code/modules/mob/living/abilities/shout.dm
NanakoAC: nope, hearers is bad, it doesnt actually have anything to do with hearing

PR #14, opened 2019-12-07
Death and Damage handling

NanakoAC:
This PR fixes up necromorph death. The primary effect is to make them die when accumulating a sufficient quantity of total damage to all limbs.
Completely severing a limb makes it count more
Hits to torso parts count less

Secondly, a minor thing but it took several hours of effort. I combed through the code and removed updatehealth polling, for humans at least. Now they only update health when something actually changes, less of that horrible per-second processing

Thirdly, while testing, i noticed some horrible embedding issues, and added a no embed flag for weapons to fix this. Currently added to energy melee weapons.

Fourthly, while poking around with those i noticed and tried out the rocksaw. It was absurdly, hideously overpowered and allowed someone to singlehandedly butcher a brute. I researched the thing, and its stats here didn't at all seem to resonate with how it performs in its source material. Since this is a meaningful item for dead space, it has been nerfed to much more reasonable (but still pretty strong) levels

Lastly, fixed brutes being unkillable due to their bodyparts being too fragile to hold enough damage. they now have Giant limbs with 4x health


code/modules/mob/living/carbon/human/death.dm
```diff
@@ -110,3 +109,24 @@
...
+		var/obj/item/organ/internal/brain/brain = internal_organs_by_name[BP_BRAIN]
+		if(!brain || (brain.status & ORGAN_DEAD))
+			.= TRUE
+	if (!.)
+		.=species.handle_death_check(src)
+
+	if (.)
+		//Time to die
+		death()
+		blinded = 1
```
Kmc2000:
```suggestion
		blinded = TRUE
```

code/modules/mob/living/carbon/human/death.dm
```diff
@@ -110,3 +109,24 @@
...
+		silent = 0
```
Kmc2000:
```suggestion
		silent = FALSE
```

code/modules/organs/external/standard.dm
```diff
@@ -153,3 +153,6 @@
 	parent_organ = BP_R_ARM
 	joint = "right wrist"
 	amputation_point = "right wrist"
+
```
Kmc2000, line 156: Whitespace

code/modules/organs/internal/lungs.dm
```diff
@@ -52,13 +52,17 @@
 
 
 /obj/item/organ/internal/lungs/proc/remove_oxygen_deprivation(var/amount)
+	if (!oxygen_deprivation)
+		return 0
 	var/last_suffocation = oxygen_deprivation
 	oxygen_deprivation = min(species.total_health,max(0,oxygen_deprivation - amount))
+	if (owner)	owner.updatehealth()
```
Kmc2000: Please indent this. This looks ugly

code/modules/organs/internal/lungs.dm
```diff
@@ -52,13 +52,17 @@
...
 	return -(oxygen_deprivation - last_suffocation)
 
 /obj/item/organ/internal/lungs/proc/add_oxygen_deprivation(var/amount)
 	var/last_suffocation = oxygen_deprivation
 	oxygen_deprivation = min(species.total_health,max(0,oxygen_deprivation + amount))
+	if (owner)	owner.updatehealth()
```
Kmc2000: This too

Kmc2000 (approved):
Looks good to me.

PR #21, opened 2019-12-12
Ubermorph anims & limbs

Ketrai:
chops ubermorph into limbs, and adds regeneration animations for said limbs.


NanakoAC (approved):

NanakoAC (changes requested):
theres a bug in the l_leg regen

NanakoAC (changes requested):
Also as previously discussed, all necromorph dmis should include an empty iconstate named "blank" as the first entry. Please insert this

NanakoAC:
we have some more oddities
ubermorph testing
i took the ubermorph sprites from your PR to test with
there's a big gap in the right armpit in front view, there also seems to be several gaps around the legs there too. I suspect this is caused by not maxing out contrast on your selection when cutting up

in the left view, the right leg is entirely missing

also not a bug but subjective feedback, i can see what you were trying to do with making the left leg darker in front view, but i think you went overboard a bit, its too dark and lacking in detail. it looks fine from the back, but front is odd

the regeneration icon for the head is a little offset in the first two frames only. Those need to be moved 1 pixel to the left
frame 3 and onwards is fine, only the first two

Ketrai:
Should hopefully be fixed now (fingers crossed)

PR #48, opened 2020-01-18
Marker: Draft 1

NanakoAC:
Rough draft of marker code, ready for testing but not merging.

Main features:
-Adds the marker, with art by ketrai and modded by me, currently mapped into a temporary spot in cargo bay
-Adds the necroshop, allowing necromorphs to be spawned and inspected, this took a lot of work
-Adds signals and master, the abstract player mobs who can inhabit the body of any spawned necromorph freely. Use Join Necromorph Team in Necromorph verbs tab
-Adds the necroqueue, allowing newly spawned necros to be immediately filled with a player

The marker accumulates biomass slowly over time, and also much more quickly by absorbing corpses. Any living mob can be dragged into its vicinity and absorption starts automatically. Humans which are absorbed in this manner will be taken apart piece by piece, gradually losing limbs until they're completely gone. Spawn and drag in a big pile of corpses to test it !

The Master signal controls the marker, its currently on a "first to press the button gets to be master" basis. We should consider alternate methods in future like a poll or random selection. 

Anyone can view the necroshop, but only the master (and admins) can spend biomass to spawn things

Currently, all signals join the necroqueue automatically. In future, i should add some options to opt out of it, and to pick which necros to play


code/modules/mob/observer/freelook/marker/marker.dm
```diff
@@ -0,0 +1,148 @@
+
+/*
+	The marker is the heart of the necromorph army
+*/
+/obj/machinery/marker
```
Ketrai, line 5: Probably best to add a smaller marker as a subtype of this using the smaller sprites.

Ketrai:
1. You can't possess a necromorph that has been evacuated. It just gives you an error message saying the necromorph is already being controlled. 

2. Marker does not start to absorb bodies of those next to it once activated.  Only those dragged next to it when it's active. This also applies to humans killed in absorbing range.

3. Limbs are not seen as valid biomass.

4. Absorption happens rather fast. Would probably nicer if corpses are only slowly taken apart rather than in one minute or two.

5. Necromorphs can keep mutilating humans after death. Probably should be prevented if absorption gives more based off of limbs?

NanakoAC:
1. Cannot reproduce
2. fixed
3. sort of fixed, they now get deleted but don't give biomass for balance reasons
4. fixed, was a bug
5. It isn't based on limbs, working as intended

NanakoAC:
Superceded by #61 
This PR is now obsolete

PR #69, opened 2020-02-28
Minor bugfixes

boskoramen:
This allows you to pick items up (this was an issue on the dev branch, and it's because of the locs var for atoms). Also a temp fix for the compile issues with the icon paths.

I have this PR so it does not clog up my two planned PRs with actual changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ sql/test_db
 # vscode
 .vscode/*
 .history
+*.code-workspace
 
 # swap
 [._]*.s[a-v][a-z]
--- a/code/_onclick/click.dm
+++ b/code/_onclick/click.dm
@@ -94,7 +94,7 @@
 		CtrlClickOn(A, params)
 		return 1
 	if(modifiers["left"])
-		.=LeftClickOn(A, params)
+		LeftClickOn(A, params)
 		return 1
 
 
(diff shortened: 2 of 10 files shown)


code/_onclick/adjacent.dm
```diff
@@ -83,9 +83,14 @@ Quick adjacency (to turf):
 /atom/movable/Adjacent(var/atom/neighbor)
 	if(neighbor == loc) return 1
 	if(!isturf(loc)) return 0
+	var/iter_count = 0 // Checks whether locs actually contains turfs
```
NanakoAC: What exactly is the issue here? I've not noticed any problems picking up items. 
What's this code doing?

boskoramen: I’ve been running into a weird issue where the locs var contains a reference to items instead of their locs. I VVed and have a screenshot of this, but I don’t have access to that or my workspace until 3 hours from now.

This might be related to the BYOND version I’m on.

boskoramen: I’ve been running into a weird issue where the locs var contains a reference to items instead of their locs. I VVed and have a screenshot of this, but I don’t have access to that or my workspace until 3 hours from now.

This might be related to the BYOND version I’m on.

NanakoAC: Please find out which version that is. If its a byond bug, we need to report it.

Snypehunter007: Is there any updates on this?

PR #95, opened 2020-03-01
Brute and Door balancing

NanakoAC:
A package of changes to the necromorph brute, and to various hard targets around the ship.
Brute:
-The brute is now effectively three tiles wide while using its charge attack. It collides appropriately with mobs and dense objects either side of it while charging. This means the charge can no longer be used in a confined space - need a corridor at least 3 wide, but it is also much harder to dodge.
-Added new brute sprites, including the fleshy variant. Variant has identical stats, just different visuals, 50% chance to be randomly selected
-Slightly reduced the cooldown on the brute's slam attack
-Slightly increased the brute's speed while charging
-Improved structure damage bonus on brute punch

Hard Targets:
-Majorly reduced the health of all doors, generally to about half of what they were before. Their damage resistance is unchanged, so this tweak more significantly affects heavy hitters like brute.
-Reduced the health of windows. Again, resistance unchanged.
-Fixed a bug where attacking mobs would be frozen in place during the opening animation when breaking a door.

Miscellaneous Changes:
-Charge impacts now apply impulse to mobs, knocking them away if the charger is heavier than the victim
-Fixed a bug where structure damage factor wasn't being applied to unarmed attacks against doors, making those hits weaker than they should have been


Ketrai:
Can you still punch other brutes to propel them? Large necros should probabaly be exempt from being knocked around by most things.

Ketrai:
Also, you can make it so closing a pr will automatically close issues you have open. I don't exactly remember the syntax though. 

https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

PR #96, opened 2020-03-02
Slasher Dodge and Leaper Fixes

NanakoAC:
A package of tweaks aimed at rebalancing the Slasher necromorph a bit

Misc: 
Base biomass cost reduced, from 80 to 70

Defense:
-Max health reduced from 80 to 70
-Base evasion removed entirely
-Added a new active ability, Dodge. When used, it makes the slasher jump one tile to the side, and gain a massive boost to evasion for 1.5 seconds. With proper timing they can evade big bursts of damage, and especially heavy weapons. 6 second Cooldown. Dodge is bound to altclick, charge moved to ctrl+alt+click

Charge Attack:
Slightly increased speed of charge
Slightly reduced windup time of charge
Massively increased duration of charge, so it will chase targets for longer.

Generally, these charge tweaks will make it tricky to escape a slasher in an open space, you'll need to turn a corner and break line of sight so he crashes into something

Leaper and other changes:
Leaper Tweaks:
	The leaper has its new sprites, including per-limb sprites, and it can now be dismembered.
	The leaper's tail is now an excellent weapon to pick up and throw after severing it

	Tailstrike:
		Slightly reduced the damage of tailstrike, and increased cooldown
		
	Leap:
		Increased the flight speed of leap
		Reduced the windup time of leap
		Reduced the punishment stun time for missing with a leap
	
	Gallop:
		Adds a new Gallop ability, gives the leaper a huge and loud burst of speed for 4 seconds, allowing it to clatter through hallways at a terrifying pace.
		Based off their behaviour in dead space 2.
		Gallop comes with a drawback of making the leaper more vulnerable. If they hit by something, or bump into a dense object while galloping, they take some damage and get knocked down for a while. Using it in combat or in tight spaces is not recommended
		
		
	Misc:
		Fixed an issue where leaper could spam tailstrike to make noise
		Fixed several critical bugs that were making all melee weapons unable to hit things
		Fixed a lack of functionality that made it impossible to target a leaper's tail
		Fixed a runtime error in limb damage calculations for severed limbs


Ketrai:
Noting thw potential of the force gun to toss a stack of leaper tails at necros as a hyper efficient bomb of sorts. As such is it possible to make the leaper tail break after one use? Same should be implemented with glass shards. There ueed to be an exploit on cm where grenades could launch a stack of glass shards and instantly kill many xenos.

PR #204, opened 2020-03-30
Puker

NanakoAC:
Hooray! Its been a long time coming

This PR implements the puker, a Tier II / Elite ranged necromorph.  With art by Ketrai, and audio ripped by jacob danik, selected and edited by myself
Puker is fairly tough, powerful at all ranges, and flexible. He is slow moving though, and his melee attack is weak. His main strength is advanced acid-based ranged attacks;

Vomit: The signature ability, this has to be seen
https://i.imgur.com/yS43fBh.gif
Projects a massive stream of corrosive bile over a wide area, coating anyone caught in it with acid.
Secondly, it coats the ground too, creating acid patches that will splash on anyone who walks in them. These patches can grow larger with repeated vomit applications, and will gradually dry up over time. Walking in them has a unique footstep noise.
Vomit can be aimed by moving the mouse while using it, nowhere nearby is safe
Vomit has a 2 second windup time, with some audio warnings to encourage people to get away. It lasts for 3 seconds once it starts, and the user can't move while vomiting (but they can turn on the spot)

I'm really proud of how the decals turned out

Snapshot & Longshot: Identical to the spitter attacks of the same name, except that these versions deal significantly more damage

Corrosive Vengeance: 
Whenever a limb is severed, a spurt of acid blasts in all directions, making the puker dangerous in melee

Eyeless Horror: 
Puker is not blinded by losing its head. However, the vomit attack loses some range without a head

Additional Features:
All acid now slows the victims. This applies to all the puker attacks, spitter attacks, and cysts

Cysts now have audio for firing. Oddly, the cyst audio was in the puker files

Minor tweaks to limb health to make decapitation more possible

Technical Stuff
Fixed some CSS errors in the necroshop

Implemented a brand new particle system of my own design. The existing SS13 effect system was an ugly half-implemented mess, not even worth fixing

Moved various acid reagents into their own unified file

other stuff...


Ketrai:
I hope to deer god the puker doesn't ignore walls though. I'll do some tests

PR #221, opened 2020-04-09
Signal sprite randomization

Ketrai:
Literally just adds 14 more signal sprites that are randomly given to signals as they generate. Later-on people might be able to unlock certain signal sprites, or choose  one they like in general.


Ketrai:
25 sprites in total now

PR #224, opened 2020-04-11
Signal Abilities: Phase 1

### LINKED ISSUE #121: SUGGESTION: Signals and Resources - Making Necro Engaging
Hello! Let's get started. My main big concern here is "fun" for the Necromorph side, and to potentially streamline current issues the Necromorph mechanics currently have, into a way that is more balanced and fun for the whole team. Ready? Okay!

**The Problem:** The Master Signal has too much control, and far too much to juggle, leaving the smaller Marker players very little to do besides screaming for spawns. The Resource acquiring mechanics I believe can also be better shifted.

**The Suggestions:** 
1. Signals and Resources - Every Signal player should have their own resource pool, for the time being, i'm thinking this should be mostly a slow static gain, that potentially will get higher throughout the round based on how far the Corruption has spread. (Ideally Corruption spread will be slower.) 
     Signals can spend their own resources on "Basic" Necromorphs, each with different costs. Somewhat like their own smaller store, but very limited. Things such as Slashers, Pukers, Leapers, etc. These Necromorphs can either be spawned at either The Marker itself, a Nest, or a ventilation shaft for a higher cost, for ambushes. In addition, i'm thinking Signals should be able to effect "Insane" players with voices and flickering lights perhaps, to kill time.

2. The Master Signal - On the other hand, the Master Signal will have their own main pool of Resources, that much like Blob in normal SS13, increases in gain and capacity from the amount of Corruption on the Map. Either way, the Master Signal spends their Resources on primarily spreading the Corruption, structures, and "Advanced" Necros. (Enhanced Types, Guardians, Brutes, Infectors, etc.)

**Benefits of this Slight Rework:** First of all, it solves the problem of relying too heavily on Corpses, making Infectors actually not wasting resources by converting people when they are added. (Recycling corpses can still be used for a temporary "Bonus" boost of Resources for the master signal, but it's not their sole source anymore. Meaning player Pop isn't the source of Necros)

It takes some of the overwhelming stress off of the Master player from having a full team of 8 other Signals yelling at them to spawn Slashers, and if they decide to spend it on a Brute instead, the cost is simply the "Team" can't play since all their points was put into one Necro.

It makes Necromorph Signal gameplay more engaging, with people having to manage their own small point pools, without making them too overpowered, and giving more player choice. Do they want to opt for staying and observing? Maybe messing with insane people for a slightly stronger Puker? Or do they just want to keep spamming out as a weak Slasher?

I think it would make it easier to control and balance without it all being reliant on a single point pool, a single player, and make it much more of a team effort to do things. Potential problem may result in "Bloat" if too many people join the Necromorph side, perhaps resources could be shared between the team, or there can only be so many "Signal Players" depending on corruption size to begin with. (So you don't result in 20 Necro players and 5 Human players for example.) As well, if Slashers are too strong, you can adjust their cost or strength individually from a player choice perspective.

**Conclusion:** This Idea is probably not perfect, there might be some exploits or oversights i'm overlooking, but I genuinelly think it's a good base or skeleton or at least a concept to build off of, and would make Necromorph gameplay much more engaging in the long term. Make it more of a team effort instead of "Master Signal Please Do Everything while I am along for the ride"

That Concludes Masked's Rambling. Feedback is welcome.

### LINKED ISSUE #129: SUGGESTION: Choice conflict for necromorph players.
#### Description of issue
There's an incentive for the marker to save up for more expensive units, but signals just want the marker to spawn more cheap necros quick so they can play. There's also an issue where the marker can't really spend money on structures as it prolongs waiting times.

NanakoAC:
A big project, that's taken most of the past two weeks. This PR adds special abilities useable by signals and markers.

Each of those players gets a reserve of Psi Energy, which ticks up slowly, and is preserved through mob changes. Both types can store 15 minutes worth of energy generation. The marker is always gaining energy, while the signal players only gain energy while they are a signal (ie, not when controlling a necromorph)

Psi Energy can be spent on a (for now) modest range of abilities that have minor roleplay and gameplay effects. Several new corruption nodes have been added which are spawned by energy, mostly weaker versions of existing ones. Many of these spells can only be used after the marker becomes active, but a few can be used before.

Two of these spells are for marker use only. Reposess lets it kick a player out of their vessel, a long requested feature to deal with AFKs. And Assault Wave provides a global buff with a very long cooldown, an experiment in pacing necromorph tactics into coordinated attack waves. We'll see how things pan out.

This PR represents about 25% of the total abilites i'd planned to add, so it is far from a finished feature. But there's enough content here to provide some fun gameplay right now

Many of the future ability plans have other dependancies (like psychosis and AI), but many of them also could be done right now if we had the time. It's just the case that this feature has taken long enough and i'd like to get it into the next test

And as my healing touch spreads across the codebase, this also fixes a bunch of bugs i ran into along the way which got lost in the haze. mostly oddities related to corruption spread and signal registration. I also mass-replaced maxHealth to proper snake case max_health across the code, which accounts for about half of the file changes, so its not as big as it appears

closes #121 as it addresses a lot of those ideas
closes #129 too, as there are now alternate ways to do many of the features of corruption


code/_helpers/lists.dm
```diff
@@ -627,7 +629,7 @@ proc/dd_sortedTextList(list/incoming)
 /datum/alarm/dd_SortValue()
 	return "[sanitize_old(last_name)]"
 
-#define subtypesof(prototype) (typesof(prototype) - prototype)
+
```
NanakoAC, line 632: moved elsewhere

PR #293, opened 2020-04-29
nerfs exploder reclamation time

Ketrai:
3 minutes => 5.

Mainly to make them less spammable, fully upgraded nests pretty much make 1 exploder per 2 minutes, meaning that with two nests you'd already have an exploder each minute, that's a bit insane.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/necromorph/exploder.dm
+++ b/code/modules/mob/living/carbon/human/species/necromorph/exploder.dm
@@ -13,7 +13,7 @@
 	biomass = 40
 	mass = 50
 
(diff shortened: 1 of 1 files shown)


NanakoAC:
hmm, not sure about this, their low reclamation time is a defining feature, how about i nerf the explosion strength instead ?

NanakoAC:
thinking about this a bit more, its a step in the right direction, but may be too much. Balance should be done in small increments

go with 4 minutes for now, we can see how that goes in combination with other upcoming explosion fixes, and then see if it still needs increased more after that

PR #324, opened 2020-05-03
Bugfixes/Tweaks

sleepysquidd:
- Removes thunderdome network from station network list
- Only logs head position jobs instead of all jobs
- Fixes multiz stairs

### DIFF
--- a/code/modules/multiz/structures.dm
+++ b/code/modules/multiz/structures.dm
@@ -215,7 +215,7 @@
 			above.ChangeTurf(/turf/simulated/open)
 	. = ..()
 
-/obj/structure/stairs/Uncross(atom/movable/A)
+/obj/structure/stairs/Exit(atom/movable/A)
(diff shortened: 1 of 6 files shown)


code/modules/multiz/structures.dm
```diff
@@ -215,7 +215,7 @@
 			above.ChangeTurf(/turf/simulated/open)
 	. = ..()
 
-/obj/structure/stairs/Uncross(atom/movable/A)
+/obj/structure/stairs/Exit(atom/movable/A)
 	if(A.dir == dir && upperStep(A.loc))
 		// This is hackish but whatever.
```
NanakoAC, line 220: is this still hackish?

sleepysquidd: Yeah kinda, I _think_ the new Exit proc means we could drop the if statement & upperStep proc

PR #325, opened 2020-05-03
Bugfixes and some formatting changes

boskoramen:
Fixes pushing objects out of range from retaining grab.

Fixes climbing ladders with pulled atoms not working.

Fixes elseif not working in TComms scripting.

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -21,6 +21,7 @@
 	var/mob/pulledby = null
 	var/item_state = null // Used to specify the item state for the on-mob overlays.
 	var/does_spin = TRUE // Does the atom spin when thrown (of course it does :P)
+	var/mid_diag_move = FALSE // Whether the atom is in the first step of diagonal movement
 
 
 	//Mass is measured in kilograms. It should never be zero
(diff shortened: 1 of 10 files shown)


code/game/atoms_movable.dm
```diff
@@ -21,6 +21,7 @@
 	var/mob/pulledby = null
 	var/item_state = null // Used to specify the item state for the on-mob overlays.
 	var/does_spin = TRUE // Does the atom spin when thrown (of course it does :P)
+	var/mid_diag_move = FALSE // Whether the atom is in the first step of diagonal movement
```
NanakoAC, line 24: This seems like a hacky way to do this. 
Have you looked at any other codebases which implement diagonal movement?

NanakoAC:
This contains a lot more than your description indicates, including some high risk changes to movement. We need more explanation on these

code/game/atoms_movable.dm
boskoramen: We discussed this in Discord. To reiterate, this is necessary because diagonal movement is converted to a horizontal and vertical movement. If we want diagonal movement to actually just be a single motion, we'd need to incorporate appropriate checks for dense atoms to prevent people from moving through two objects diagonally.

PR #328, opened 2020-05-04
Tool power consumption fixes

Toxici11i:
This is separate from the saw fix for now(in hindsight it may break something) so its better to separate.

### DIFF
--- a/code/game/objects/items/weapons/tools/saws.dm
+++ b/code/game/objects/items/weapons/tools/saws.dm
@@ -95,10 +95,10 @@
 
 
 /obj/item/weapon/tool/saw/plasma/turn_on()
-	.=..()
-	flick("plasma_saw_ignite", src)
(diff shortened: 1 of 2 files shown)


NanakoAC:
as discussed, the saw change should be included in this, not seperate (its not up in a PR at all)

code/game/objects/items/weapons/tools/saws.dm
```diff
@@ -95,11 +95,11 @@
 
 
 /obj/item/weapon/tool/saw/plasma/turn_on()
-	.=..()
-	flick("plasma_saw_ignite", src)
-	playsound(get_turf(src), 'sound/weapons/saberon.ogg', 20, 1, -2)
-
+	.=.()
```
NanakoAC: this needs to be .=..()

i'm not sure why it compiles like this, but a dot has been deleted and it won't work right

NanakoAC:
approved

NanakoAC (approved):

PR #329, opened 2020-05-04
Bye infinite saws!

Toxici11i:
Saws no longer break thermodynamics and work without power, or a cell.


Toxici11i:
closing, fixed in another PR

Issue #348, opened 2020-05-10
[Bug] Necros can catch viruses


TheMaskedMan2:
Title

ghost:
Fixed with PR #358

PR #358, opened 2020-05-14
hopefully this may fix ISSUE #348

ArcadeDan:


ArcadeDan (commented):
added virus_immune to each necromorph variation

PR #365, opened 2020-05-15
Ladders Stronk

### LINKED ISSUE #319: [Bug] Puker Acid Melts Ladders
This is bad because ladders are very important. Title.

Afya212:
Makes ladders indestructible.
Fixes #319

### DIFF
--- a/code/modules/multiz/structures.dm
+++ b/code/modules/multiz/structures.dm
@@ -9,6 +9,8 @@
 	desc = "A ladder. You can climb it up and down."
 	icon_state = "ladder01"
 	icon = 'icons/obj/structures.dmi'
+	breakable = FALSE
+	unacidable = 1
 	density = 0
 	opacity = 0
 	anchored = 1


NanakoAC (commented):
Have you tested this? I significantly overhauled structure damage

NanakoAC:
there are also broken ladder sprites in the same file as the existing ones.  a bit incomplete though (missing floor/ceiling hatch versions). with a little work on photoshop that could be done, and we could have ladders enter a broken state without deleting themselves, from which they could be repaired

Feel up to doing that? or too far beyond scope

Afya212:
Forgot to test puker, seems that it was an old issue, ree. I always test my code before making a PR of course. I'd still argue it's useful in same way it is with stairs, having them blow up can just ruin flow and make it far more inconvenient. 

Spriting is one thing I don't touch at all though, beyond scope.

NanakoAC (approved):
seems fine to implement as-is

PR #366, opened 2020-05-15
Exploder Changes

### LINKED ISSUE #302: Exploders can destroy plating
#### Description of issue
exploders can destroy plating, making holes in the floor or space. this is a huge issue, as it'll kill anyone not in a suit, and make people drop down levels.

#### Difference between expected and actual behavior
for exploders to not be complete bio-terrorists

#### Steps to reproduce 
be exploder, go boomboom

Afya212:
Optional Balance PR:
Reduces the chance of exploder breaching ceiling tiles, increases bio explosion.

Resolves #302

### DIFF
--- a/code/modules/mob/living/carbon/human/species/necromorph/exploder.dm
+++ b/code/modules/mob/living/carbon/human/species/necromorph/exploder.dm
@@ -190,13 +190,13 @@ The last resort. The exploder screams and shakes violently for 3 seconds, before
 	spawn()
 		bioblast(epicentre = T,
 		power = 60,
-		maxrange = 4,
+		maxrange = 5,
 		falloff_factor = 0.3)
 
 	//A normal explosion
 	spawn()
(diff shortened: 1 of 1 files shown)


NanakoAC:
this isn't minor, thats a very heavy nerf there
The linked issue was already mostly addressed by removing the highest explosion tier from them.  I didn't close it after the last explosion PR. 
Farther nerfing in hopes of reducing chances are not a good solution

best fix would be to revamp floor code. I wrote some good floor code on eris with multiple layers.  It shouldn't be possible to open a hole to space in the surface of aegis VII

There is also a bug snype pointed out, that explosions ex_act things on adjacent zlevels even when they fail to break through. That should be changed; require a hole

Afya212:
Yeah I get ya, it was a potential change, but like I said just a minor thing. I'm not touching floor code right now though. x3

PR #385, opened 2020-05-18
Makes Slasher more HUMAN-like

Eluxor:
### Adds a new Slasher spriter replacing old one.

dreamseeker_ANg7OhVxPn
dtec5OvqvI
dreamseeker_bDTnviIwoH

I feel like this could be a bit complicated, but I belive the new sprites would let players be able to see slashers more easily, they are more human like, their limbs are more visible and when removed they affect the sprite a bit more. It would make players of both factions be able to see what LIMB exactly a Slasher lost.

I just want to have an open discussion of pros and cons of a design that its just made to help other players. The sprite is not final and can be changed, palette wise, size wise and style wise. As I said before I am open to all interpretations.
I just feel like that It would improve the gameplay. 

image

-Eluxor


spicyfly:
At this moment it is important that our current sprites coincide with each other, our slasher is compliant in this sense. If you would like to assist in our endeavor please contact a member of the development team to see where your skills can be utilized.  Thanks!

NanakoAC:
As I mentioned yesterday, if you wish to make new slasher sprites, they must conform to the existing skeleton. Changing the overall size and shape is not an option

spicyfly:
Eluxor you should probably update this

PR #481, opened 2020-07-02
Makes the marker gamemode end when most of the human crew are dead

Kmc2000:
Title. Requested by West. I have no real way to test this by myself and with my being unable to select any job locally.

Tests I ran:
-Spawn in, kill self as BST. 
-Spawn in braindead mob, kill braindead mob

This basically just looks at the crew manifest whenever someone dies, and determines whether most of the players are dead, if so, end the round as a necro win.

:cl:Kmc2000
rscadd: The marker gamemode will now end when the majority of the human crew are dead.
/:cl:

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -38,6 +38,10 @@ var/global/list/additional_antag_types = list()
 	var/waittime_l = 60 SECONDS				 // Lower bound on time before start of shift report
 	var/waittime_h = 180 SECONDS		     // Upper bounds on time before start of shift report
 
+	//Used for modes which end when everyone dies.
+	var/list/dead_players = list()
+	var/player_count = 0
+
 /datum/game_mode/New()
 	..()
 	// Enforce some formatting.
@@ -49,6 +53,7 @@ var/global/list/additional_antag_types = list()
 		latejoin_antag_tags = antag_tags.Copy()
 	else if(!round_autoantag && latejoin_antag_tags.len)
 		round_autoantag = TRUE
+	player_count = GLOB.all_crew_records.len
 
 /datum/game_mode/Topic(href, href_list[])
 	if(..())
@@ -541,3 +546,19 @@ proc/get_nt_opposed()
 			to_chat(usr, "[ticker.mode.extended_round_description]")
 	else
 		to_chat(usr, "<i>Shhhh</i>. It's a secret.")
+
+//Handlers for people dying / being revived.
+/datum/game_mode/proc/on_crew_death(mob/living/carbon/human/H)
+	if(!H)
+		return
+	player_count = GLOB.all_crew_records.len
+	dead_players += H
+	check_finished()
+
+/datum/game_mode/proc/on_crew_revive(mob/living/carbon/human/H)
+	if(!H)
+		return
+	player_count = GLOB.all_crew_records.len
+	for(var/mob/M in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
+		if(M == H)
+			dead_players -= H
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -549,16 +549,22 @@ proc/get_nt_opposed()
 
 //Handlers for people dying / being revived.
 /datum/game_mode/proc/on_crew_death(mob/living/carbon/human/H)
-	if(!H)
-		return
+	if(!H?.client)
+		return FALSE
+	for(var/X in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
+		if(X == H.client.ckey)
+			return FALSE
 	player_count = GLOB.all_crew_records.len
-	dead_players += H
+	dead_players += H.client.ckey
 	check_finished()
+	return TRUE
 
 /datum/game_mode/proc/on_crew_revive(mob/living/carbon/human/H)
-	if(!H)
-		return
+	if(!H?.client)
+		return FALSE
 	player_count = GLOB.all_crew_records.len
-	for(var/mob/M in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
-		if(M == H)
+	for(var/X in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
+		if(X == H.client.ckey)
 			dead_players -= H
+	check_finished()
+	return TRUE
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -549,22 +549,22 @@ proc/get_nt_opposed()
 
 //Handlers for people dying / being revived.
 /datum/game_mode/proc/on_crew_death(mob/living/carbon/human/H)
-	if(!H?.client)
+	if(!H?.mind)
 		return FALSE
-	for(var/X in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
-		if(X == H.client.ckey)
+	for(var/datum/mind/M in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
+		if(M == H.mind)
 			return FALSE
 	player_count = GLOB.all_crew_records.len
-	dead_players += H.client.ckey
+	dead_players += H.mind
 	check_finished()
 	return TRUE
 
 /datum/game_mode/proc/on_crew_revive(mob/living/carbon/human/H)
-	if(!H?.client)
+	if(!H?.mind)
 		return FALSE
 	player_count = GLOB.all_crew_records.len
-	for(var/X in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
-		if(X == H.client.ckey)
-			dead_players -= H
+	for(var/datum/mind/M in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
+		if(M == H.mind)
+			dead_players -= H.mind
 	check_finished()
 	return TRUE
--- a/code/game/gamemodes/marker/gamemodes.dm
+++ b/code/game/gamemodes/marker/gamemodes.dm
@@ -48,8 +48,7 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
 	var/marker_setup_time = 45 MINUTES
 	var/marker_active = FALSE
 	antag_scaling_coeff = 8
-	var/minimum_alive_percentage = 5 //How much grace do you want to give the necros for killing people? The default value for this is 5. If we have a crew of 10 people, then you have to kill 9 of them to win as necro.
-	var/dead_players = 0
+	var/minimum_alive_percentage = 10 //How much grace do you want to give the necros for killing people? The default value for this is 5. If we have a crew of 10 people, then you have to kill 9 of them to win as necro.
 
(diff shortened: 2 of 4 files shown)


code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -123,4 +118,29 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
...
+
+*/
+
+//Handlers to end the game if everybody dies.
+/mob/living/carbon/human/death(gibbed,deathmessage="seizes up and falls limp...", show_dead_message = "You have died.")
+	. = ..()
+	var/datum/computer_file/report/crew_record/R = get_crewmember_record(real_name) //Try get a crew manifest for this mob
+	if(!R)
+		return FALSE //Not on the manifest? You don't exist.
+	if(istype(ticker.mode, /datum/game_mode/marker)) //Precondition: Ensure that the game mode is one of ours, we don't want to say, override traitor or anything.
```
NanakoAC: this isnt a great way to do things. If code exists for a specific gamemode, it should be called -on- that gamemode, add a proc to all gamemodes to notify them that a crewmember has died, and let the gamemode handle things from there

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -123,4 +118,29 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
...
+		ticker.mode.check_finished()
+
+//Method to be called when a human player dies for any reason, this will check the amount of players that should be alive, and sees if they're dead or not. We're doing it via an almost signal (bay lacks signals) to avoid processing overhead.
+/datum/game_mode/marker/check_finished()
+	var/all_mobs = GLOB.all_crew_records.len //If you're not in the manifest I don't care about your existence. Updates this every time we check a win to ensure that no new players have joined the fray.
+	dead_players ++
```
NanakoAC: Don't change variables within a checking function. And don't have a simple number for dead players at all.

make it a list - of real names or mind datums, which can be checked

NanakoAC (changes requested):
Needs a little work, see comments

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -51,6 +48,8 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
 	var/marker_setup_time = 45 MINUTES
 	var/marker_active = FALSE
 	antag_scaling_coeff = 8
+	var/minimum_alive_percentage = 5 //How much grace do you want to give the necros for killing people? The default value for this is 5. If we have a crew of 10 people, then you have to kill 9 of them to win as necro.
```
NanakoAC: make this value 10%

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -123,4 +118,29 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
...
+	. = ..()
+	var/datum/computer_file/report/crew_record/R = get_crewmember_record(real_name) //Try get a crew manifest for this mob
+	if(!R)
+		return FALSE //Not on the manifest? You don't exist.
+	if(istype(ticker.mode, /datum/game_mode/marker)) //Precondition: Ensure that the game mode is one of ours, we don't want to say, override traitor or anything.
+		ticker.mode.check_finished()
+
+//Method to be called when a human player dies for any reason, this will check the amount of players that should be alive, and sees if they're dead or not. We're doing it via an almost signal (bay lacks signals) to avoid processing overhead.
+/datum/game_mode/marker/check_finished()
+	var/all_mobs = GLOB.all_crew_records.len //If you're not in the manifest I don't care about your existence. Updates this every time we check a win to ensure that no new players have joined the fray.
```
NanakoAC: Put a check here to see if the marker has activated yet. We don't want the round to end because one player spawned in alone, and shot themselves. This victory condition should only be able to activate after the necromorphs come into play

code/game/gamemodes/game_mode.dm
```diff
@@ -541,3 +546,19 @@ proc/get_nt_opposed()
 			to_chat(usr, "[ticker.mode.extended_round_description]")
 	else
 		to_chat(usr, "<i>Shhhh</i>. It's a secret.")
+
+//Handlers for people dying / being revived.
+/datum/game_mode/proc/on_crew_death(mob/living/carbon/human/H)
+	if(!H)
+		return
+	player_count = GLOB.all_crew_records.len
+	dead_players += H
```
NanakoAC: unfortunately this won't work, human corpses can be deleted in explosions or flying out of the map, and definitely will be deleted in the far-more-likely-scenario of being eaten by the marker.

The dead players list should store mind datums please, those represent a character

also, rename the list to dead crewmembers or dead crew

code/game/gamemodes/game_mode.dm
```diff
@@ -541,3 +546,19 @@ proc/get_nt_opposed()
...
+	check_finished()
```
NanakoAC, line 559: check finished should just be in the marker gamemode's override of this function

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -123,4 +117,20 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
...
+/**
+
+There is probably a better way to do this, but I've added some more stringent checks to the gamemode win conditions to prevent things like:
+Admin characters being counted as crew, and on their deletion, ending the game
+Non-critical characters like any ghost-roles you may wish to add, or even antags, from counting as "dead crew" (Antags fallback to their own "are the antags dead checks"
+
+*/
+
+//Method to be called when a human player dies for any reason, this will check the amount of players that should be alive, and sees if they're dead or not. We're doing it via an almost signal (bay lacks signals) to avoid processing overhead.
+/datum/game_mode/marker/check_finished()
```
NanakoAC, line 131: i dont see the base class defininition of this function, did it already exist?

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -123,4 +117,20 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
...
+	if(!marker_active)
+		return ..()  //Change this to FALSE if you want to override all other win conditions until the marker activates.
+	if(dead_players.len >= round(player_count-(player_count*((minimum_alive_percentage > 0) ? minimum_alive_percentage/100 : 0)))) //Allows the necromorphs a bit of grace so they don't have to hunt down _everyone_, as that could lead to cheese. The ? statement is there to handle a config value of 0% grace, if you decide to go for that.
+		return TRUE //Necros win!
```
NanakoAC, line 135: check finished returns true, then what? how does this actually cause the round to end?

this is being called from on_crew_death, which doesn't do anything with the return value

NanakoAC (changes requested):
more work needed

code/game/gamemodes/marker/gamemodes.dm
Kmc2000: It's a standard gamemode proc called on tick to check if the mode should end or not, when check finished returns TRUE, the round ends (to my understanding)

code/game/gamemodes/game_mode.dm
Kmc2000: My main issue with using minds, is that when that player leaves, the mind will be nulled out. So if everyone left the game then necros win regardless of whether they should've. I could probably store it as a list of ckeys (text) which would solve both those issues

NanakoAC: I'm about 80% certain that minds are not deleted when someone leaves, are you sure? 
They persist between connections because they're tied to an IC character, not a client. Most servers use them for cloning

A list of ckeys would not work, since its possible for a player to respawn and play multiple characters in one round.

Minds, crew records, or character real names. I'm pretty confident minds are the one to go with

NanakoAC (changes requested):
the list of dead players needs to use minds, or real names. The mob itself can be deleted so those are no good, crew records can also be deleted so best not to use them either, and names are nonunique so there is some potential for collisions

code/game/gamemodes/game_mode.dm
Kmc2000: Minds are 100% deleted when you leave, at least in my experience with /tg/ code. Not having a mind is tantamount to being AFK

code/game/gamemodes/game_mode.dm
```diff
@@ -541,3 +546,25 @@ proc/get_nt_opposed()
 			to_chat(usr, "[ticker.mode.extended_round_description]")
 	else
 		to_chat(usr, "<i>Shhhh</i>. It's a secret.")
+
+//Handlers for people dying / being revived.
+/datum/game_mode/proc/on_crew_death(mob/living/carbon/human/H)
+	if(!H?.client)
```
NanakoAC: i'm not sure this would even compile with that questionmark, presumably a typo
But regardless, there's no reason that clientless mobs shouldn't count. People can disconnect and be killed while not connected

code/game/gamemodes/game_mode.dm
```diff
@@ -541,3 +546,25 @@ proc/get_nt_opposed()
...
+		return FALSE
+	for(var/X in dead_players) //Hacky, yes. But more consistent than locate() in whatever.
```
NanakoAC: ckeys aren't the right thing to use, a person can respawn and play more than one character during a round. Other things can take control of a body in some cases

Minds or Real Names or Crew records, ideally minds

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -123,4 +117,20 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
...
+Admin characters being counted as crew, and on their deletion, ending the game
+Non-critical characters like any ghost-roles you may wish to add, or even antags, from counting as "dead crew" (Antags fallback to their own "are the antags dead checks"
+
+*/
+
+//Method to be called when a human player dies for any reason, this will check the amount of players that should be alive, and sees if they're dead or not. We're doing it via an almost signal (bay lacks signals) to avoid processing overhead.
+/datum/game_mode/marker/check_finished()
+	if(!marker_active)
+		return ..()  //Change this to FALSE if you want to override all other win conditions until the marker activates.
+	if(dead_players.len >= round(player_count-(player_count*((minimum_alive_percentage > 0) ? minimum_alive_percentage/100 : 0)))) //Allows the necromorphs a bit of grace so they don't have to hunt down _everyone_, as that could lead to cheese. The ? statement is there to handle a config value of 0% grace, if you decide to go for that.
```
NanakoAC, line 134: one more thing, put a minimum on this, like requiring at least 5 crew total over the round, otherwise it will interfere with debugging 

and maybe neaten up this line a bit

NanakoAC (changes requested):
The dead players list is still not right. Minds!

code/game/gamemodes/game_mode.dm
Kmc2000: The questionmark is entirely intentional, to check that H exists as well as H's client.

PR #493, opened 2020-07-07
Necro night vision, probably..?

Kmc2000:
For some reason the see_in_dark var is extremely inconsistent in my copy, but I can't find anything else that changes it...so?

### DIFF
--- a/code/modules/mob/living/carbon/human/species/necromorph/necromorph.dm
+++ b/code/modules/mob/living/carbon/human/species/necromorph/necromorph.dm
@@ -171,11 +171,20 @@
 /datum/species/necromorph/get_icobase(var/mob/living/carbon/human/H)
 	return icon_template //We don't need to duplicate the same dmi path twice
 
+/mob/proc/night_vision()
+	set name = "Toggle night-vision"
+	set desc = "This doesn't actually turn up anywhere, so hi!"
+	set category = SPECIES_NECROMORPH
+	var/toggled = (see_in_dark > 0)
+	to_chat(src, "<span class='notice'>You [toggled ? "disable" : "enable"] your night-vision.</span>")
+	set_see_in_dark( (toggled) ? 0 : 100 )
+
 /datum/species/necromorph/add_inherent_verbs(var/mob/living/carbon/human/H)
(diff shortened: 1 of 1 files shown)


NanakoAC (changes requested):
Completely on the wrong track here, see_in_dark is native byond behaviour that has nothing to do with our coded-in-DM lighting system.

The nightvision we need exists in the code somewhere. Play as a ghost and see how it works, as you fly around dark areas you should notice they have a little square around them, about 3x3 tiles in size, and darkness under it is lessened. 

That's the system we need to use. Figure out what it is, where the code is for it, how it works, and how to control it so we can set it on necromorphs and change its range

Kmc2000:
> Completely on the wrong track here, see_in_dark is native byond behaviour that has nothing to do with our coded-in-DM lighting system.
> 
> The nightvision we need exists in the code somewhere. Play as a ghost and see how it works, as you fly around dark areas you should notice they have a little square around them, about 3x3 tiles in size, and darkness under it is lessened.
> 
> That's the system we need to use. Figure out what it is, where the code is for it, how it works, and how to control it so we can set it on necromorphs and change its range

https://github.com/DS-13-Dev-Team/DS13/blob/1fd277620f07b1d7eccf01ba47eca5e0066c66b7/code/modules/mob/living/silicon/robot/life.dm#L283
https://github.com/DS-13-Dev-Team/DS13/blob/0472aa0fe988209c14bf9b0352300a86761bdbbb/code/datums/observation/see_in_dark_set.dm#L24
https://github.com/DS-13-Dev-Team/DS13/blob/0295d2b503aaa5abcb8310c08f08665cf3eabeaa/code/modules/mob/observer/ghost/ghost.dm#L37
?

NanakoAC:
http://www.byond.com/docs/ref/#/mob/var/see_in_dark

Its native behaviour and it does essentially nothing for this use case.
It has some kind of marginal effect on view distance but it doesnt affect our lighting overlays

what ghosts have (in addition) is something different, figure out what

NanakoAC:
This PR is obsolete, the feature has been implemented

PR #504, opened 2020-07-13
Fixes webmap again

AffectedArc07:
Stuff brok


NanakoAC (changes requested):
How exactly does this fix the webmap ? All you seem to have done is slightly rearrange our global lists

And not with any consistency, you made one neater and one messier, in exactly the same manner.
These were syntactically valid before, why these changes?

AffectedArc07:
The tool which renders the webmaps isn't native DM so it has to do its best to parse together DM code into something it can read. In this case, defines have to be on one line (Hence moving the `list(` up to the same line.

You're well within your right to close this, you just won't get functioning webmaps.

NanakoAC:
but this doesn't put defines on one line. It doesn't even consistently put the first element of the list on the first line

Wouldn't it be better to tweak the tool's parsing engine to handle these cases ? The byond compiler can do it

AffectedArc07:
Parsing DM in another language is difficult enough, having the define arguments be split over multiple lines is a pain and not something **I** can do.

NanakoAC:
Why this file though? Why does a webmap tool need to look anywhere except the .dmm files?

NanakoAC:
My main concern here is, how are we going to maintain this?  If something like this can break the webmap, then it WILL happen again unintentionally. Are you going to fix it for us every time ?

I don't have any way of checking it to find out what's wrong

AffectedArc07:
It has to scan the entire codebase to build up an object tree. A DMM file is just a dictionary of objects, which is then referenced by the code. The DMM has no clue what a `turf/simulated` is, let alone the texture.


Also given that I have fixed it before, and now, and it's a 2 line change, it seems fine for me to make fixes for this.

NanakoAC (approved):
then this seems fine

PR #558, opened 2020-07-25
Kinesis

NanakoAC:
Implements kinesis, comes standard on engineering rigs, also added to loot lists. 

comes with audio and visual FX
overhauls throw code, adds cool new pixel movement
Adds Spawn Bluespace Engineer admin verb, creates a BST with an engineering rig, good for testing
restores modded tools to loot list
organises a lot of helper files along the way

Press F to activate, click and drag objects to manipulate them, Press F while holding an object to launch it away from you.
Uses a fair bit of physics calculations, kinetic objects will maintain velocity when you let them go, and keep flying for a bit

Dragging objects into mobs is mildly useful, actually launching them is much more powerful

Engage throw mode and drag an object into yourself to grab it out of the air

Power costs are very low, since it will be needed for objective specific things in future


JakeDaBoss:
Will test out later today.

JakeDaBoss (changes requested):
Few changes I'd like to request:

Kinesis can currently only be enabled with hotkey mode enabled. For those who prefer not to use hotkey mode (such as myself), make a verb for it under the hardsuit model.

Bugs:
I've been throwing things at Necromorphs, but they do not seem to connect to them, or damage them. Granted, these were simplemobs that weren't controlled.

Snypehunter007:
We also don't need another button to spawn a slightly different Bluespace Tech.

JakeDaBoss (changes requested):
A list of possible bugs:

* Items are able to be merged into walls by Kinesis. Makes for some visually annoying glitches (in my experience). http://prntscr.com/tphcv0
* If dismembering a Slasher, and their left/right leg flies off, the sprite does not update to reflect this. http://prntscr.com/tphe4w (Note: It works for arms and head. Legs remain as they are, even when cut off.)
* Ability to lift heavy objects such as lockers. Personally, I feel like this may be a bit too much, though that is up to interpretation.
* Can lift things through glass. http://prntscr.com/tphgi7 http://prntscr.com/tphgry
* Using throw or any kind of inventory actions activates telekinesis.
* Rare occurence: Sometimes placing items back down in an irregular position (http://prntscr.com/tphjjg) can cause smooth movement to go tits up, and makes you jump tiles.


Currently all I could find.

NanakoAC:
1. Not fixable. Not without establishing a really complex framework to figure out the extents of every sprite in the game
2. Unrelated, this is an issue with dismemberment masking (or rather, the lack of it) and applied to humans too before this PR. It'll be a bit of an effort to fix and is not caused by kinesis
3. Everything is movable if its not anchored down, moving heavy objects uses physics based on mass. However there is an issue that lockers aren't accounting for their contents to determine their mass, correcting it....

NanakoAC:
4. bug, fixing
5. Not clear what you mean. Kinesis is a toggled mode, you turn it on to throw stuff around and turn it off when you're done throwing stuff. Its definitely going to interfere with normal gameplay if you leave it on at all times, that's why its bound to a hotkey for easy access
6. Can't reproduce, not sure what is meant here. However its worth pointing out that objects inherit velocity, it will keep moving for a second after being dropped if it hasn't come to a complete stop

NanakoAC:
RE: Hotkey mode, i will add a verb for it, but its going to be somewhat inconvenient to click that
RE: Things not connecting: The smooth movement is a visual effect, collision is still based on direct tile entry. Aim for the centre of the target's tile to hit them. There is a bit of skill to it, but that's not a bad thing
RE: Bluespace engineer: I need it, and there's no reason to argue with more debug tools. The bluespace engineer comes with a rig, the BST doesn't, this gives quick access to two mobs with different levels of defense that i can use for future combat testing.  Every debugging tool i make shaves a few precious seconds off routines that i do thousands of times

NanakoAC:
Latest Commit adds proper mass checking to lockers, crates and storage objects;
Adds a statpanel verb for toggling kinesis
Partially fixes grabbing through windows. An unrelated bug in projectile code means that they can still be grabbed through single-pane windows, but they can't be pulled through glass
 this will be fixed in another PR at a later date, as it will require a significant reworking of projectile code, and is beyond scope of this PR

PR #564, opened 2020-07-25
Fixes the Admin System.

JakeDaBoss:
Adminhelps are now received again.

Autoresponse also removed from adminpm.


JakeDaBoss:
Please do not merge until @Snypehunter007 or @WestHybrid confirms they've replaced the current admin_ranks.txt

PR #567, opened 2020-07-26
Deleted Content Restoration

NanakoAC:
Restores several changes that were made in the past month without adequate consultation:

-Loot Distribution: Restores many random loot spawns, not as many as before, and not quite as disorganised.
-Loot spawners now have a smaller, less intrusive icon so they don't look messy in the map editor
-Loot spawners now use multiple items in one spawn, so that there are no longer large clumps of spawners on each tile
-Swaps the ore boxes for wooden crates in the marker room, so they can hold stuff
-Restores armor pieces to fire closets, they're the closest readily available thing to an outfit closet, and already contain certain sorts of protective gear naturally
-Restores the lobby screen code (but not the images, so we're still only using the single hand). The code will support multiple screens in future
-Slightly tweaks rare loot distribution for more power nodes and fewer rigs

kinesis PR should be merged before this one


Snypehunter007:
I can't comment on the first three points, I'd have to see what they look like, but the most immediate thing, to me, are the fifth and sixth changes.

Multiple problems with armor in fire lockers, which already have been presented before.

A.) The most populous departments, Engineering, Security, and Mining already are provided RIG suits for their jobs. They are not going to trade out for different things.
B.) Not everyone needs a piece of armor on the ship.
C.) It makes no sense for their to be a random piece of armor in every fire locker, it is a locker with protective equipment to fight fires, not contain a random piece of armor. It'd be like the equivalent of giving welding supplies and spare metal or glass in every O2 locker. On top of that, there are so many fire lockers around the ship that it's ridiculous how many pieces of armor we'd be supplying to the crew.
D.) It is providing armors we do not want in the codebase as part of the round.

For the lobby change, we don't need it. It is really that simple. No reason to rotate screens (especially when we have no other ones) and it doesn't really need to be set up that way at all in the first place. If you want to show multiple pieces of artwork, just set it up so when a new round starts, it picks from one of the available lobby-arts available (in the future if we get more).

PR #594, opened 2020-08-07
Kinesis Window Fixes

NanakoAC:
Fixes another of the three major bugs

I found a way to do this by only modifying the invisible test projectiles, normal projectile behaviour is unchanged. Would still like to overhaul it in future, but its a bit complex

Also some general polish to kinesis targeting and FX

changes: 
  - rscadd: "Kinesis now does visual and audio effects for attempted grabs which don't find something to pickup. These failed grabs cost extra power."
  - tweak: "Kinesis power costs increased a moderate degree"
  - bugfix: "Kinesis can no longer grab objects through windows and walls."

### DIFF
--- a/code/_helpers/check_trajectory.dm
+++ b/code/_helpers/check_trajectory.dm
@@ -85,6 +85,38 @@
 	return results //Send it back to the gun!
 
 
+//Version optimised for mass testing, and verbose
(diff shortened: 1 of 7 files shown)


JakeDaBoss:
Will test this out prior to testing today.

PR #597, opened 2020-08-07
MapFixes&CodeAdjustmentsV.2

Snypehunter007:
- Added a air alarm to the Processing Room.
- Added a pistol to the Captain's safe, as a last resort.
- Added medical wall lockers to the Aegis Colony.
- Added the Kellion Technician outfit to the code.
- Added a standard device power cell to the Supply Officer's locker to power the Stick of Discipline.
- Added six more Planet Cracker slots. Due to space constraints, six of the Planet Crackers voidsuit storage units are on Aegis, this will be temporary.
- Added another Cargo Serviceman slot.
- Assigned Line Cook and Bartender only accesses and added them to the appropriate doors.
- Changed the Captain's safe to a modern, wall-safe.
- The Monitoring Room door bolt button now works.
- Fixed the door name on the janitorial closet, south of Cargo.
- Fixed the windows on the Aegis Colony Security Checkpoint to be the correct type of windows.
- Fixed the Kellion Engineer spawning with Unitologist gear.


code/datums/outfits/spec_op.dm
```diff
@@ -193,21 +209,21 @@
...
+/decl/hierarchy/outfit/kendra
+	name = "Kellion Technician"
+	uniform = /obj/item/clothing/under/kellion/tech
+	suit = /obj/item/clothing/suit/storage/kellion_jacket
 	l_ear = null
-	mask = /obj/item/clothing/mask/breath
-	head = /obj/item/clothing/head/helmet/space/unitologist/deacon
-	belt = /obj/item/weapon/storage/belt/utility/full
-	back = /obj/item/weapon/rig/vintage
+	mask = null
```
JakeDaBoss, line 217: Just remove what is not applicable. Keeps everything clearer.

JakeDaBoss (changes requested):

code/datums/outfits/spec_op.dm
Snypehunter007: That is true, I suppose. The reason I left the null stuff in is to keep each entry in the same order (for the list of slots) instead of being a mishmash of different slots.

IDK, maybe I should just keep a commented out example of the order at the top?

JakeDaBoss: Nanako is against it, but I'm hugely in favor for documentation needed at a later date.

PR #632, opened 2020-08-10
ERT Telecomm Room

sleepysquidd:
Adds a TComm room to the USM Valor for ERT teams


Snypehunter007 (changes requested):
Err, it might be best to make the room non-interactable by normal mobs.

I'm sorry, I should have mentioned it before. It might be better to move it to a non-breachable room near the cargo elevator and Thunderdome.

Sorry again about this Squidd.

Snypehunter007:
Thinking on it more, I think this is fine actually. I'll change my opinion to approved and merge this now.

Good job, Squidd.

PR #652, opened 2020-08-11
Ert's

Awantje:
Todo

- [x] Place Landmarks for spawns
- [x]  Review my own code in the morning
- [x]  Check text on spawn in.


code/game/antagonist/outsider/ert.dm
```diff
@@ -1,42 +1,117 @@
...
-
+	valid_species = list(SPECIES_HUMAN)
 	flags = ANTAG_OVERRIDE_JOB | ANTAG_SET_APPEARANCE | ANTAG_HAS_LEADER | ANTAG_CHOOSE_NAME | ANTAG_RANDOM_EXCEPTED
 	antaghud_indicator = "hudloyalist"
-
-	hard_cap = 5
-	hard_cap_round = 7
-	initial_spawn_req = 5
-	initial_spawn_target = 7
+	hard_cap = 99
```
Awantje: More todo: afaik these are not used at all. Look into their removal.

Awantje: Removed

Snypehunter007 (changes requested):
For the change-log, I'd say you could put everything on one line:

  - rscadd: "Added a working ERT system, with three new Response Teams now available."

And then the change log file name should be:

Name-PR-(PR #).yml

So, for example:

Snypehunter007-PR-648.yml

Beyond that, I think it looks pretty good outside of that flavor text stuff we talked about, which is fine because we don't currently have text to replace those with.

Snypehunter007 (approved):
Thanks!

PR #654, opened 2020-08-11
Movement Optimisation

NanakoAC:
A major step in reducing movement lag:

This PR gives each turf an internal list of things it contains which could possibly block movement, based on a variable set on atoms.

Only things in that list are checked for collisions.

This saves at least one check on every tile by not trying to collide with lighting overlays. It saves many more in cases of pipes, wires, catwalks, corruption, decals etc. Things which are in every room.
And it saves vast quantities of checks in cluttered areas littered with trash, bodyparts, blood, et al.

Basically, floors that don't contain dense stuff no longer generate collision events at all

We had 48 million calls to canpass in the most recent data, this will help

Also supplemental change: Explosions no longer break power cables. This was an unfun feature to have

### DIFF
--- a/code/game/atoms.dm
+++ b/code/game/atoms.dm
@@ -31,6 +31,19 @@
 	var/default_alpha = 255
 	var/default_scale = 1
 
+
+
+	/*
+		OPTIMISATION
+			If set false, this atom will never be checked to recieve collisions. Everything will move through it as if it isnt there and canpass will never be called.
+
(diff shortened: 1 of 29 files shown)


AshTheGuy:
Based?

AshTheGuy:
> Based?

Dont think its based at all, shit now lags 100x worse. think you messed up a line while coding.

PR #688, opened 2020-08-18
Changes the credits song to the DS2 credits

aguyiguess:
Changes the credits song to the dead space 2 end credits as well as removing it from the lobby.


JakeDaBoss:
I'd prefer this gets closed. The song is changed on a local branch for me, but also with some major overhauls.

aguyiguess:
fair point, carry on gamer

PR #713, opened 2020-08-23
Admin Communicate to +ADMIN.

JakeDaBoss:


NanakoAC (commented):
why though, its useful to moderators and devs too, its a way to speak in necrochat without joining and leaving

PR #755, opened 2020-08-28
Ingested/injected acid now hurts the liver (and kidneys)

aguyiguess:
Damage dealt to the liver is the acid power. 
Sulphuric: 2.5 power
Hydrochloric: 3 power
Polytrinic: 5 power
Biological(necromorph): 3.5 power

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/acid.dm
+++ b/code/modules/reagents/Chemistry-Reagents/acid.dm
@@ -11,6 +11,10 @@
 
 /datum/reagent/acid/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	M.take_organ_damage(0, removed * power)
+	if(ishuman(M))
+		var/mob/living/carbon/human/H = M
+		var/obj/item/organ/internal/liver/liver = H.internal_organs_by_name[BP_LIVER]
+		liver.take_general_damage(power)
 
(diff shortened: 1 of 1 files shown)


code/modules/reagents/Chemistry-Reagents/acid.dm
```diff
@@ -11,6 +11,10 @@
 
 /datum/reagent/acid/affect_blood(var/mob/living/carbon/M, var/alien, var/removed)
 	M.take_organ_damage(0, removed * power)
+	if(ishuman(M))
```
NanakoAC, line 14: any operations with reagents need to factor in the removed value. Damage should be removed*power

What's the goal with this though? Injecting acid is already pretty dangerous, this is going to make it essentially an instakill. It wouldn't affect necromorphs but it would allow some really fast human vs human killing, it seems excessive. Acid is generally meant for external applications, and poisons for internal

even removed*power is going to be a huge quantity of damage, and essentially instakill with a few units

NanakoAC (commented):
a code issue, and more general concerns of balance and purpose

code/modules/reagents/Chemistry-Reagents/acid.dm
aguyiguess: I injected 15u of polyacid into someone before and he didn't come close to dying, at least I think. though I do see how making acid essentially a 2nd poison would be excessive. I feel like acid should do _some_ organ damage though, would it be possible to do removed * power *.2 or .3 to reduce the damage?

PR #800, opened 2020-09-01
Fixes2

### LINKED ISSUE #488: [Suggestion] Put an oxygen recycler in the escape pods and solitary cells of brig
Body (for suggestions):
Suggestion:
An oxygen recycler that automatically turns carbon dioxide into oxygen

What do you think it'd add:
Currently if you escape via a escape pod and wait around 5-7 minutes you will start being affected by carbon dioxide, which will give you some red messages (you feel dizzy) and then it will kill you after inhaling too much carbon dioxide.

This also happens with the solitary cells of brig, they have no vents and no scrubbers so the same things happens, but it takes more time, as they have 4 tiles instead of 2 (pods).

### LINKED ISSUE #670: [Suggestion] Add a cryo pod room (like brig) for SSD crew. (Or a cryo room in upper deck)
#### Suggestion:
Currently, over 90% of the SSD Personnel end up and medbay, and the cryo pods are simply too far to get the SSD guy in the cryo pod room, so I thought we could have a room or a few scattered cryo pods like brig has to put SSD personnel inside.

Or we could also add a cryo pod room in the upper deck.

Cryo pods in medbay would be time efficient and allow SSD people to unlock their job slot, same with a cryo pod room in upper deck

JakeDaBoss:
Closes #488 
Closes #670


code/modules/necromorph/corruption/cyst.dm
```diff
@@ -17,12 +17,12 @@
 	plane = EFFECTS_BELOW_LIGHTING_PLANE //Cysts need to draw over many things
 	var/obj/item/ammo_casing/cystbomb/payload = null
 
-	biomass = 5
+	biomass = 15
 	biomass_reclamation = 1
-	reclamation_time = 5 MINUTES
+	reclamation_time = 7.5 MINUTES
```
NanakoAC: These aren't minor changes, maybe in the right direction but excessive. Do this instead:
Biomass cost: 8
reclamation change is fine
recharge delay, leave it as is, 5 seconds is precisely based on the games, that won't ever change

JakeDaBoss: 5 seconds is too short and too spammy. Especially done if spammed out the ass.

NanakoAC: Cysts have a lot of weaknesses. 
-Any cyst caught in their blast (including themselves) is destroyed. 
-You can trigger their firing by tossing something infront of them. Chuck a lighter or an empty mag to create safe passage
-Their limited arc of fire makes them easy to flank
-Their bombs can be picked up with kinesis and throw back, or at necros.
-They have a range limit of 6, meaning anyone with any gun can stand just at the edge of their range and snipe them

they're an area denial tool but far from invulnerable, they can kill quickly if someone walks into a trap, thats working as intended. Cysts have recently been buffed significantly because they were a joke before, their current state isn't far off right

JakeDaBoss: You do realize 5 seconds is nothing right? You can't walk 2 feet without it firing again.

Also, you want authenticity? Power to you. But authenticity vs gameplay effects have to be put into consideration, and you're not doing a whole lot by trying to defend their current state. You turned them from a joke to Tripod-worthy.

NanakoAC: let us worry about design. i know the cyst is a bit overtuned and i'll take care of it, so just remove all the cyst changes and let me handle it x

NanakoAC:
aside from cyst changes, code looks good. pending mapping approval from snype

code/modules/necromorph/corruption/cyst.dm
JakeDaBoss: I've changed it to 6.5. I implore you to try this out at least. Cysts are now spammed out the ass, and adjustments have to be made.

Snypehunter007 (approved):
I'm good with the mapping changes.

Gets my +1 for those.

**EDIT:** I didn't look at the cryo changes.

code/modules/necromorph/corruption/cyst.dm
NanakoAC: adjustments are needed, but not like this. 
Let me handle it, you don't have to worry about it. Just remove all cyst related changes, i'll make a PR today

NanakoAC (changes requested):
revert cyst changes, we'll handle them seperately

PR #813, opened 2020-09-03
Wheelchair speed fix

Terror4000rus:
Right now you can use a wheelchair to speed up your character without any drawbacks from items or wounds.
I "fixed" it without new checks and other changes. I think it's better than see a fast train of crew with a man in wheelchair at it's head or even have a `nice time` to catch a man in it without admin's help.


NanakoAC (approved):
its a simple solution but it works for now

PR #818, opened 2020-09-05
Evac fixes

JeserTheFox:
- bugfix: "Fixed evacuation called by admins not being forced." 
  - rscdel: "Disabled evacuation cancelling code due to being outdated and not doing anything." 
  - rscadd: "Added admin verb Delay Evacuation that allows to increase or decrease time until evacuation is allowed." 
  - rscadd: "Added admin warnings that trigger at approximately 10 and 5 minutes until evacuation will be allowed normally." 

Nanako needs to check and, most likely, fix evac points formulas, since they misbehave.

### DIFF
--- a/code/controllers/evacuation/evacuation_lifepods.dm
+++ b/code/controllers/evacuation/evacuation_lifepods.dm
@@ -21,8 +21,8 @@
 		return list()
 	if (is_idle())
 		return list(evacuation_options[EVAC_OPT_ABANDON_SHIP])
-	if (is_evacuating())
-		return list(evacuation_options[EVAC_OPT_CANCEL_ABANDON_SHIP])
+//	if (is_evacuating())	//disabled until proper cancel evac is coded
+//		return list(evacuation_options[EVAC_OPT_CANCEL_ABANDON_SHIP])
 
 #undef EVAC_OPT_ABANDON_SHIP
 #undef EVAC_OPT_CANCEL_ABANDON_SHIP
--- a/code/controllers/evacuation/evacuation_lifepods.dm
+++ b/code/controllers/evacuation/evacuation_lifepods.dm
@@ -23,6 +23,7 @@
 		return list(evacuation_options[EVAC_OPT_ABANDON_SHIP])
 //	if (is_evacuating())	//disabled until proper cancel evac is coded
 //		return list(evacuation_options[EVAC_OPT_CANCEL_ABANDON_SHIP])
+	return list()
 
 #undef EVAC_OPT_ABANDON_SHIP
 #undef EVAC_OPT_CANCEL_ABANDON_SHIP
--- a/code/controllers/evacuation/evacuation_pods.dm
+++ b/code/controllers/evacuation/evacuation_pods.dm
@@ -62,9 +62,9 @@
 		return list()
 	if (is_idle())
 		return list(evacuation_options[EVAC_OPT_ABANDON_SHIP])
-	if (is_evacuating())
-		if (emergency_evacuation)
-			return list(evacuation_options[EVAC_OPT_CANCEL_ABANDON_SHIP])
+	//if (is_evacuating())
+		//if (emergency_evacuation)	//disabled until proper cancel evac is coded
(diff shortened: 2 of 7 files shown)


code/controllers/evacuation/evacuation_lifepods.dm
```diff
@@ -21,8 +21,8 @@
 		return list()
 	if (is_idle())
 		return list(evacuation_options[EVAC_OPT_ABANDON_SHIP])
-	if (is_evacuating())
-		return list(evacuation_options[EVAC_OPT_CANCEL_ABANDON_SHIP])
+//	if (is_evacuating())	//disabled until proper cancel evac is coded
```
NanakoAC, line 24: This should probably return an empty list if none of the conditions are met, and make sure whatever recieves that list can safely handle empty values

code/controllers/evacuation/evacuation_pods.dm
```diff
@@ -62,9 +62,9 @@
 		return list()
 	if (is_idle())
 		return list(evacuation_options[EVAC_OPT_ABANDON_SHIP])
-	if (is_evacuating())
-		if (emergency_evacuation)
-			return list(evacuation_options[EVAC_OPT_CANCEL_ABANDON_SHIP])
+	//if (is_evacuating())
```
NanakoAC, line 65: as above, for all these files

code/game/gamemodes/marker/evac_points.dm
```diff
@@ -32,12 +39,37 @@
...
 		//No addtimer call here, so we'll stop gaining points now
 	else
+		check_admin_warnings()
 		pointgain_timer = addtimer(CALLBACK(src, .proc/charge_evac_points), 1 MINUTE, TIMER_STOPPABLE) //Shuttle was unable to be called. Try again recursively.
 	return FALSE
 
-
+//proc that handles messages to admins regarding close evac time
+/datum/game_mode/marker/proc/check_admin_warnings()
+	var/time_till_evac = (evac_threshold - evac_points) * minutes_per_point MINUTES	//for Nanako, this formula here needs checking
```
NanakoAC: You made a proc for this elsewhere, use that instead of pasting code around

code/game/gamemodes/marker/evac_points.dm
```diff
@@ -32,12 +39,37 @@
...
+	if(time_till_evac <= 10 MINUTES)
+		if(show_admin_warning_10)
+			show_admin_warning_10 = FALSE
+			message_admins("Approximate time until evacuation is unlocked is less than 10 minutes.")
+		else
+			if(time_till_evac <= 5 MINUTES && show_admin_warning_5)
+				show_admin_warning_5 = FALSE
+				message_admins("Approximate time until evacuation is unlocked is less than 5 minutes.")
+
+/datum/game_mode/marker/proc/get_time_until_evac()
```
NanakoAC, line 59: Every usage of this function just multiplies it by MINUTES (which converts it to deciseconds) so you should do that here too instead of at the caller.

All time in byond is handled in deciseconds, best to stay consistent

NanakoAC (commented):
Mostly fine, just some minor tweaks

code/controllers/evacuation/evacuation_lifepods.dm
JeserTheFox: Returns empty list() by default now. Only called in comms module of consoles, it handles empty lists just fine.

code/controllers/evacuation/evacuation_pods.dm
JeserTheFox: Returns empty list() by default now. Only called in comms module of consoles, it handles empty lists just fine.

code/game/gamemodes/marker/evac_points.dm
JeserTheFox: Replaced with get_time_until_evac()

code/game/gamemodes/marker/evac_points.dm
JeserTheFox: Added MINUTES to this proc, removed from places where it's called.

PR #820, opened 2020-09-06
Chem dispenser no longer requires the Complex Devices skill

### LINKED ISSUE #804: Suggetion and might be bug: chemistry and my sanity
#### Description of issue: chemistry machines doing what they shouldn't, delivering broken numbers of chems, delivering other chems besides the one i asked it to send a example is inaprovaline, you add acetone, carbon and sugar what i got the first time is a broken mess between -2 and +2 of the amount i wanted plus other chemicals, the first time it added silicon which made kelotane, the second time i made received the plus sulfuric acid
the chem master does the same

it doesn't do what it should do, give broken mess of number instead of actual amount it was asked for

#### Steps to reproduce:
do chemistry

#### Suggestion:
give the chemistry machine the usual which is infinite chems but limited by power
and IF the bug is not a bug remove that "content" and fix chemistry
make my life easier and my sanity ok and medical job less painful and wasteful

aguyiguess:
Closes #804
Tested, it works

### DIFF
--- a/code/modules/reagents/dispenser/dispenser2.dm
+++ b/code/modules/reagents/dispenser/dispenser2.dm
@@ -151,17 +151,7 @@
 		var/label = href_list["dispense"]
 		if(cartridges[label] && container && container.is_open_container())
 			var/obj/item/weapon/reagent_containers/chem_disp_cartridge/C = cartridges[label]
-			var/mult = 1 + (-0.5 + round(rand(), 0.1))*(user.skill_fail_chance(core_skill, 0.3, SKILL_ADEPT))
-			C.reagents.trans_to(container, amount*mult)
-			var/contaminants_left = rand(0, max(SKILL_ADEPT - user.get_skill_value(core_skill), 0)) * can_contaminate
-			var/choices = cartridges.Copy()
-			while(length(choices) && contaminants_left)
-				var/chosen_label = pick_n_take(choices)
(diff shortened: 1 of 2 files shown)


code/modules/reagents/dispenser/dispenser2.dm
```diff
@@ -151,17 +151,7 @@
 		var/label = href_list["dispense"]
 		if(cartridges[label] && container && container.is_open_container())
 			var/obj/item/weapon/reagent_containers/chem_disp_cartridge/C = cartridges[label]
-			var/mult = 1 + (-0.5 + round(rand(), 0.1))*(user.skill_fail_chance(core_skill, 0.3, SKILL_ADEPT))
-			C.reagents.trans_to(container, amount*mult)
-			var/contaminants_left = rand(0, max(SKILL_ADEPT - user.get_skill_value(core_skill), 0)) * can_contaminate
```
JakeDaBoss, line 156: Fully against this.

People without the complex devices (it is a complex device) skill can now go mad zoomies on the machine and never fuck up.

A: Not realistic.
B: Defeats purpose of doctors.

aguyiguess: Although it isn't realistic, I don't think anyone actually likes this feature at all. It just isn't fun to click "All" 7 times when you want to make something simply because you forgot to set Complex Devices to 3.

PR #825, opened 2020-09-06
Wall frames color fix

Terror4000rus:
Fixed color of wall frames. Not hull one (because a mapper have to do so).

### DIFF
--- a/code/game/objects/structures/wall_frame.dm
+++ b/code/game/objects/structures/wall_frame.dm
@@ -136,9 +136,6 @@
 /obj/structure/wall_frame/hull
 	color = COLOR_HULL
 
-/obj/structure/wall_frame/prepainted
-	color = COLOR_GUNMETAL
-
 /obj/structure/wall_frame/hull/Initialize()
 	. = ..()
 	if(prob(40))
--- a/code/game/objects/structures/wall_frame.dm
+++ b/code/game/objects/structures/wall_frame.dm
@@ -12,7 +12,6 @@
 	density = 1
 	throwpass = 1
 	layer = TABLE_LAYER
-	color = COLOR_GUNMETAL
 
 	var/damage = 0
 	max_health = 10
@@ -136,6 +135,9 @@
 
 /obj/structure/wall_frame/hull
 	color = COLOR_HULL
+
+/obj/structure/wall_frame/prepainted
+	color = COLOR_GUNMETAL
 
 /obj/structure/wall_frame/hull/Initialize()
 	. = ..()
--- a/code/game/objects/structures/wall_frame.dm
+++ b/code/game/objects/structures/wall_frame.dm
@@ -12,6 +12,7 @@
 	density = 1
 	throwpass = 1
 	layer = TABLE_LAYER
+	color = COLOR_GRAY
 
 	var/damage = 0
 	max_health = 10
--- a/code/modules/projectiles/ammunition.dm
+++ b/code/modules/projectiles/ammunition.dm
@@ -165,8 +165,6 @@
 	if(!stored_ammo.len)
 		to_chat(user, "<span class='notice'>[src] is already empty!</span>")
 		return
-	if(alert("Do you want to empty the magazine?",,"Yes","No") == "No")
-		return
 	to_chat(user, "<span class='notice'>You empty [src].</span>")
 	for(var/obj/item/ammo_casing/C in stored_ammo)
 		C.forceMove(user.loc)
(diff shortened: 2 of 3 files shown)


code/game/objects/structures/wall_frame.dm
```diff
@@ -137,6 +136,9 @@
 /obj/structure/wall_frame/hull
 	color = COLOR_HULL
 
+/obj/structure/wall_frame/prepainted
+	color = COLOR_GUNMETAL
```
JakeDaBoss: Why is this necessary? Wallframes aren't meant to be colored. They do not act like walls in the sense that they should be painted.

Also, we don't use painted versions of the wall currently.

ghost (commented):
If you can make this aspect of your fix a separate PR, we can get it merged, the other one may need a further look at from the team.

code/game/objects/structures/wall_frame.dm
Terror4000rus: And that's why I made this subtype. If someone want to use prepainted walls, they will use this. If not - previous type (which is already used at the ship and colony, but has own color while walls don't).

JakeDaBoss: We won't ever use it. Ishimura won't use it, neither will Tau. Sprawl is a long shot for similar reasons. It seems you added it because you think/hope they'll be used, but I can tell you now, they won't. I've spoken to Snype about this multiple times.

Terror4000rus: It looks a little dumb. You get mad about one new suitable type (because there already other type with same color in code) and hold the whole PR because of it. Good mappers have their own vision on their job and may see situations where current color will be better than regular blank walls.

Terror4000rus:
> If you can make this aspect of your fix a separate PR, we can get it merged, the other one may need a further look at from the team.

...Alright. I'll make other branch for magazines tweak. This is so ... , really.

code/game/objects/structures/wall_frame.dm
JakeDaBoss: I'm not mad, I'm saying it's wasted effort and time.

Terror4000rus:
I will add examples of poor mapping and poor mapping ahead, just to end this pointless conversation conversation.
Upd: wrong, I don't have a reason to up the server than for it.

code/game/objects/structures/wall_frame.dm
```diff
@@ -137,6 +136,9 @@
 /obj/structure/wall_frame/hull
 	color = COLOR_HULL
 
+/obj/structure/wall_frame/prepainted
```
JakeDaBoss: Not needed. The rest is a decent addition.

JakeDaBoss (changes requested):

code/game/objects/structures/wall_frame.dm
Terror4000rus: ...You already said this opinion before. People aren't so stupid to don't understand what you wanted to say about it.

JakeDaBoss: The previous was a comment, this was me actually opposing it. There's a way GitHub works, you know.

Terror4000rus:
I appreciate my time to don't hold horns, @WestHybrid 
Just hope mappers will have more skill than one who made the ship with such wall frames. The bridge windows still make me sick.

Snypehunter007:
@Terror4000rus if you have mapping suggestions, other than the two you've already made, feel free to post them here on the Github.

Also, just to make sure I'm on the same page, what exactly does this change do?

JakeDaBoss:
> @Terror4000rus if you have mapping suggestions, other than the two you've already made, feel free to post them here on the Github.
> 
> Also, just to make sure I'm on the same page, what exactly does this change do?

Removes color from the current wallframes. Which should be promptly denied, unless we want discoloration with respect to walls.

Terror4000rus:
Sorry, but are you blind?

Terror4000rus:
I rechecked and something strange appeared, but now, it has to be better.
Before:
<details>

image


</details>
After:
<details>

image


<details>

NanakoAC:
I'm not quite seeing the difference in the above pictures, can you explain a bit more whats going on here?

Terror4000rus:
> I'm not quite seeing the difference in the above pictures, can you explain a bit more whats going on here?

Wall frames in lower picture have almost the same color as walls. One above - don't.

NanakoAC (approved):
i see now, its subtle, but good change

PR #834, opened 2020-09-08
Removing stuck from crit failure effects

Terror4000rus:
image
Nice joke
I got my BOTH arms broken in same round because of it.


NanakoAC (approved):
We'll accept this as a temporary measure

these things will return, reworked in future with an improved skill system

PR #836, opened 2020-09-08
Rebranding NTNet to SolNet

aguyiguess:
There are a LOT of NTNet files in game so I might've missed something, there was at least 75 files to go through.

NTNet is referred to as SolNet in-game now. Not any of the variables or procs. 

Additionally, in the laptop_vendor.dm file, I changed NT-branded to Zebra Industries.


Googles-Hands:
An bit late, but wouldnt "EGNet" be more fitting, considering the major political power is the EarthGov or "CECNet" since it could be seen as an company owned system

PR #850, opened 2020-09-12
Balances tweaks and fixes v3

JakeDaBoss:


JakeDaBoss:
This is DIRTY AS FUCK. And probably REALLY bad practice. But... It works! :tm: I think.

PR #857, opened 2020-09-13
On-going work.

JakeDaBoss:
This branch is on-going work to address some mild/minor problems.


JakeDaBoss:
The branch name is a literal waste of effort. I could only think of one thing to do today.

JakeDaBoss:
Reserved for further bugfixing.

ghost (approved):

ghost (approved):
This along with the Activate Marker changes are approved.

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -116,10 +116,14 @@ GLOBAL_DATUM_INIT(shipsystem, /datum/ship_subsystems, new)
 	set category = "Admin"
 	set desc = "Forces the marker to immediately activate"
 
-	var/confirm = alert(src, "You will be activating the marker. Are you super duper sure?", "Make us Whole?", "Send in the Necromorphs!", "On second thought, maybe not...")
-	if(confirm != "Send in the Necromorphs!")
+	if(check_rights(R_MOD))
```
NanakoAC, line 119: this is already an admin verb, you don't need additional rights checks in here. they're handled before the verb is even added to a mob

code/modules/reagents/Chemistry-Reagents/acid.dm
```diff
@@ -6,8 +6,8 @@
 	color = "#db5008"
 	metabolism = REM * 2
 	touch_met = 50 // It's acid!
-	var/power = 2.5
-	var/meltdose = 10 // How much is needed to melt
+	var/power = 1
```
NanakoAC, line 9: these are very large changes, why so?  try to do smaller increments

code/game/gamemodes/marker/gamemodes.dm
JakeDaBoss: Fairly sure it's not. It's shown to Mentors.

code/modules/reagents/Chemistry-Reagents/acid.dm
JakeDaBoss: I've tested these changes extensively. Even something as 1.5 will still put someone in immediate soft-crit.

60 units is still enough to put someone into permanent hardcrit and passed out state from pain/shock.

JakeDaBoss: I just now noticed that you meant the other acids in specific.

So, Triflic Acid is still the biggest one available, but it's only natural that lesser Acids, such as sulphuric or whatever have less power.

The original was way too strong.

JakeDaBoss:
In general, this is a first round of absolute fixing for Triflic, and related acids. If it proves to be too ineffective, I'll raise it again.

code/game/gamemodes/marker/gamemodes.dm
NanakoAC: then its in the wrong admin category, correct the category. rights checks still don't go within these functions

PR #861, opened 2020-09-14
Renames Unitologist antag to Zealot

### LINKED ISSUE #859: Suggestion: Rename 'Unitologist to 'zealot' or 'fanatic'.
-->
#### Suggestion:
As in the title, I believe the 'unitologist' antagonist role should be renamed to zealot, or fanatic, or some kind of varient.

Why do I suggest this? The church of unitology is, in-universe, a real church that, while it butts heads with Earthgov, it is an established religion that has a presence on their station, such as Titan Station seen in Dead Space 2. This helps separate those who are indeed fanatics and antagonists, and allows those who select the church of unitology to separate themselves from the, well, fanatics, or help them, if they are ICly inclined.

aguyiguess:
title
Closes #859

### DIFF
--- a/code/game/gamemodes/marker/unitologist.dm
+++ b/code/game/gamemodes/marker/unitologist.dm
@@ -10,9 +10,9 @@ GLOBAL_LIST_EMPTY(unitologists_list)
 	message_necromorphs("<span class='cult'>[usr]: [message]</span>")
 */
 /datum/antagonist/unitologist
-	role_text = "Zealot"
-	role_text_plural = "Zealots"
-	welcome_text = "You are part of a new religion which worships strange alien artifacts, believing that only through them can humanity truly transcend. You have been blessed with a psychic connection created by the <b>marker</b>, one of these artifacts. Serve the marker's will at all costs by bringing it human sacrifices and remember that its objectives come before your own..."
+	role_text = "Unitologist Zealot"
+	role_text_plural = "Unitologist Zealots"
+	welcome_text = "You are a zealot of Unitology willing to do anything in pursuit of your beliefs. You have been blessed with a psychic connection created by the <b>marker</b>, and must serve the marker's will at all costs by bringing it human sacrifices. Remember, its objectives come before your own..."
 	id = MODE_UNITOLOGIST
 	flags = ANTAG_SUSPICIOUS | ANTAG_RANDSPAWN | ANTAG_VOTABLE
 	skill_setter = /datum/antag_skill_setter/station
--- a/code/game/gamemodes/marker/unitologist.dm
+++ b/code/game/gamemodes/marker/unitologist.dm
@@ -10,8 +10,8 @@ GLOBAL_LIST_EMPTY(unitologists_list)
 	message_necromorphs("<span class='cult'>[usr]: [message]</span>")
 */
 /datum/antagonist/unitologist
-	role_text = "Unitologist"
-	role_text_plural = "Unitologists"
+	role_text = "Zealot"
+	role_text_plural = "Zealots"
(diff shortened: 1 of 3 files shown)


NanakoAC (changes requested):
the core concept of the change seems good, but the implementation is too generic. 
They are zealots OF unitology and that needs to be communicated. They aren't an entirely seperate thing from the moderates of the religion and might enjoy support from them

This PR seems to have wholly replaced one word with another, which is not the right way to do things. Generally it should say "unitologist zealot", in some cases just one or the other is appropriate

code/game/gamemodes/marker/unitologist.dm
```diff
@@ -44,20 +44,20 @@ GLOBAL_LIST_EMPTY(unitologists_list)
 
 /datum/antagonist/unitologist/proc/give_collaborators(mob/living/our_owner)
 	//our_owner.verbs |= /mob/proc/message_unitologists
-	to_chat(our_owner, "<span class='warning'>The marker has established a psychic link between you and your fellow unitologists.</span>")
+	to_chat(our_owner, "<span class='warning'>The marker has established a psychic link between you and your fellow zealots.</span>")
```
Wulfestwulf, line 47: Instead of just Zealots, how about Unitologist Zealots?

code/game/gamemodes/marker/unitologist.dm
```diff
@@ -10,8 +10,8 @@ GLOBAL_LIST_EMPTY(unitologists_list)
 	message_necromorphs("<span class='cult'>[usr]: [message]</span>")
 */
 /datum/antagonist/unitologist
-	role_text = "Unitologist"
-	role_text_plural = "Unitologists"
+	role_text = "Zealot"
```
Wulfestwulf: Unitologist Zealot instead?

code/game/gamemodes/marker/unitologist.dm
```diff
@@ -10,8 +10,8 @@ GLOBAL_LIST_EMPTY(unitologists_list)
...
+	role_text_plural = "Zealots"
```
Wulfestwulf: Unitologist Zealots instead?

code/game/gamemodes/marker/unitologist.dm
aguyiguess: that was already done, look at the latest commits

code/game/gamemodes/marker/unitologist.dm
aguyiguess: I feel like that sentence sounds better without the full name, I'm not sure about that one

code/game/gamemodes/marker/unitologist.dm
```diff
@@ -10,9 +10,9 @@ GLOBAL_LIST_EMPTY(unitologists_list)
 	message_necromorphs("<span class='cult'>[usr]: [message]</span>")
 */
 /datum/antagonist/unitologist
-	role_text = "Unitologist"
-	role_text_plural = "Unitologists"
-	welcome_text = "You are part of a new religion which worships strange alien artifacts, believing that only through them can humanity truly transcend. You have been blessed with a psychic connection created by the <b>marker</b>, one of these artifacts. Serve the marker's will at all costs by bringing it human sacrifices and remember that its objectives come before your own..."
+	role_text = "Unitologist Zealot"
+	role_text_plural = "Unitologist Zealots"
+	welcome_text = "You are a zealot of Unitology willing to do anything in pursuit of your beliefs. You have been blessed with a psychic connection created by the <b>marker</b>, and must serve the marker's will at all costs by bringing it human sacrifices. Remember, its objectives come before your own..."
```
NanakoAC: this text is less interesting than it used to be. The old version gave a bit of brief context about what unitology is and does, lets retain that

NanakoAC (approved):

PR #865, opened 2020-09-18
Divider

NanakoAC:
Adds the divider, our latest Tier II necromorph. A powerful tank who is hard to kill and excels at overwhelming a single victim. He splits into five mobs on death

Adds a bunch of peripheral improvements in the process:
-new root status and resurrection code, various other little status and effects
-Overhauls extension statmod code to greatly reduce meaningless junk. this makes coding simpler, no effect on performance.
-Lots of fixes to math, cones, lines, vector cleanup
-Proper Lesser necromorph mob implemented as a base class, used for divider components now, swarmers in future. full healthbar support
-lots of work on the wallmount code to greatly improve its support, this is used for components latching onto mobs
-reorganised movement handler code a bit, no functional changes
-Made health a property of objects instead of items, one step closer to full damage handling
-replaced bay limb breaking sounds with new ones from dead space
-Improved the execution system to support the use of weapons, paving the way for humans to do it in future
-added support for lasting damage, a new necromorph mechanic that reduces their max health. will be used for spells in future
-fixed ranged accuracy penalties for blindness and other statuses, they weren't properly transitioned to new system
-fixed crazy damage dealt to thrown mobs
-added stomping when attacking small mobs
-refactored tether code and properly named them, added damage support to them


NanakoAC:
briefly DNM while fixing some issues from internal testing

NanakoAC:
good to go, once again

PR #874, opened 2020-09-20
CodeWorkV.6

Snypehunter007:
**Code Work:**
- Reorganized our uniforms under a new category in the uniform section so that they are easier to find.
- Fixed the light RIG randomly spawning.
- Fixed the nutrient tray sprite from not showing up in Dream Maker.
- Fixed the Chief Security Officer's jacket not having armor, the name, and description.

**Sprite Work:**
- Fixed a few problems with the firesuit RIG's boot and glove sprites.
- Fixed the security uniform's east and west sprites missing two pixels and showing your stomach.


code/modules/reagents/reagent_containers/food/nutrient_blocks.dm
```diff
@@ -84,5 +84,5 @@
 /obj/item/weapon/storage/fancy/nutricube_tray/filled
 	name = "Nutricube Storage Tray"
 	desc = "A tray full of nutrient cubes, this is much heavier than it looks."
-	icon_state = "tray"
+	icon_state = "tray8"
```
JakeDaBoss: I tried fixing this before. It fixes it on the map, breaks it in the game. Could you revert this?

JakeDaBoss (changes requested):

code/modules/reagents/reagent_containers/food/nutrient_blocks.dm
Snypehunter007: What happened? I didn't notice a problem during testing.

JakeDaBoss: It was only for a round.

The nutristorage sprite is dynamically called on item removal or addition. So changing it to a hardcoded value like 8, will turn it into tray88 in game. Or tray87. Obviously those don't exist.

I have a different fix in mind, though. One that shouldn't fuck with it.

JakeDaBoss: Instead of changing it in code, just copy over a desired sprite, and call it tray in that file. Check it out in-game. Should work.

NanakoAC (changes requested):
lion is correct about the tray sprite, that needs to be fixed

PR #892, opened 2020-09-30
Re-implement old explosion behaviour

Kmc2000:
Requested by nanako. Makes some explosions handle like they used to.


Kmc2000 (commented):
Yeah I know, the define thing's weird. It didn't like me macro-ing a function itself so I've just left them there as reference for now.

PR #895, opened 2020-10-03
Necrologs.

JakeDaBoss:

### DIFF
--- a/code/modules/mob/observer/freelook/marker/signal_abilities/runes.dm
+++ b/code/modules/mob/observer/freelook/marker/signal_abilities/runes.dm
@@ -12,7 +12,7 @@
 /datum/signal_ability/runes/on_cast(var/mob/user, var/atom/target, var/list/data)
 	GLOB.cult.powerless = TRUE //Just in case. This makes sure the runes don't do anything
 	new /obj/random/rune(target)
-
+	link_necromorphs_to(SPAN_NOTICE("[user] placed a rune at LINK"), target)
 
 
 /obj/random/rune /*Better loot for away missions and salvage */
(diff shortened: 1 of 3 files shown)


code/modules/mob/observer/freelook/marker/signal_abilities/runes.dm
```diff
@@ -12,7 +12,7 @@
 /datum/signal_ability/runes/on_cast(var/mob/user, var/atom/target, var/list/data)
 	GLOB.cult.powerless = TRUE //Just in case. This makes sure the runes don't do anything
 	new /obj/random/rune(target)
-
+	link_necromorphs_to(SPAN_NOTICE("[user] placed a rune at LINK"), target)
```
NanakoAC: this would be extremely spammy, doesn't need a notification, people often mass place runes in large batches

JakeDaBoss: That's actually the entire point. It's LRP if done in mass spammings, and gives people probable cause to early metagame. This serves as an accurate way for players to ask admins for help.

NanakoAC: do we have rules about rune usage?

this is still an extremely spammy and un-optimal way to do it. if you want signals to know who made a rune, have that information stored and shown on examine

JakeDaBoss: I'll do that instead. Next week, anyway.

code/modules/mob/observer/freelook/marker/signal_abilities/runes.dm
```diff
@@ -14,10 +14,9 @@
 	new /obj/random/rune(target)
 
 
-
 /obj/random/rune /*Better loot for away missions and salvage */
 	name = "random rune"
-	desc = "This is some random loot."
```
NanakoAC, line 20: this isn't going to accomplish anything at all, i don't think it will even compile
this isn't the rune you've edited, it's the random spawner that creates a rune.
You can't have evaluated operations like [user] inside a compiletime value like desc, this data would have to be output during the examine proc. 

And of course that data would need to be put into the rune first

NanakoAC (changes requested):

PR #897, opened 2020-10-04
A fix for emotes seen by ghosts

Terror4000rus:
image
No more  things like
image
or
image


Terror4000rus:
I also could кemove prefix about the real name of an unknown spaceman at all.

JakeDaBoss (approved):

NanakoAC:
i'm not quite clear on what's being fixed here, it looks like some information is removed. can you elaborate?

Terror4000rus:
> i'm not quite clear on what's being fixed here, it looks like some information is removed. can you elaborate?

I made the message, generated for ghosts, when someone wears inappropriate ID or just hides face a bit better to look and fixed it appearance at any emotion.

PR #925, opened 2020-10-14
Chem rework.

JakeDaBoss:


code/modules/organs/internal/brain.dm
```diff
@@ -157,17 +157,19 @@
...
-				to_chat(owner, "<span class = 'notice' font size='10'><B>What's going on...?</B></span>")
-				sleep(10 SECONDS)
-				to_chat(owner, "<span class = 'notice' font size='10'><B>What happened...?</B></span>")
-			healed_threshold = 0
-
-		if(damage < (max_damage / 4))
-			healed_threshold = 1
+		if(is_necromorph())
+			return
+			if(damage > max_damage / 2 && healed_threshold)
```
NanakoAC: bad indentation has broken this entire feature :( it needs to be unindented

also necromorphs have their own version of the brain so this check shouldn't be necessary,

JakeDaBoss: They still get messages about brain damage, which is strange.

code/modules/organs/internal/brain.dm
```diff
@@ -198,6 +200,8 @@
 					damprob = owner.chem_effects[CE_STABLE] ? 30 : 60
 					if(!past_damage_threshold(2) && prob(damprob))
 						take_internal_damage(1)
+					if(can_heal)
```
NanakoAC: the brain only heals with proper blood volume and oxygenation, it seems you've tried to change that here, but i'm not clear on why. This would probably revert a recent change i made to make humans more killable

these heal checks are put right after a damage check, that's going to cause problems, i don't think they should be here.

what was the end goal here? Perhaps we can figure out a better way to achieve it

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
```diff
@@ -7,7 +7,7 @@
 	reagent_state = LIQUID
 	color = "#00bfff"
 	overdose = REAGENTS_OVERDOSE * 2
-	metabolism = REM * 0.5
```
NanakoAC, line 10: what's the purpose behind these granularly varied metabolic values? They were all the same before, that makes some math easier for medical players, is it causing problems?

code/modules/organs/internal/brain.dm
JakeDaBoss: Making sure that brain atrophy is better treatable, because as I said, brain damage is accrued way too fast, and if a person wears a rig, there's just no way they can be recovered. And everyone wears rigs.

code/modules/reagents/Chemistry-Recipes.dm
```diff
@@ -1977,4 +1977,54 @@
 	name = "Nanoblood"
 	result = /datum/reagent/nanoblood
 	required_reagents = list(/datum/reagent/dexalinp = 1, /datum/reagent/iron = 1, /datum/reagent/blood = 1)
-	result_amount = 3
\ No newline at end of file
+	result_amount = 3
+
+// New recipes as of 15th October, 2020. - Lion
+
+/datum/chemical_reaction/phlogiston
```
NanakoAC, line 1984: where did you get these from, and what do they do?

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
```diff
@@ -366,7 +366,7 @@
 	if(M.chem_doses[type] == metabolism * threshold)
 		M.confused += 2
 		M.drowsyness += 2
-	else if(M.chem_doses[type] < 2 * threshold)
+	else if(M.chem_doses[type] < 1.5 * threshold)
```
NanakoAC, line 369: rebalancing this should probably be done by editing threshold rather than the multiplier on it. can you explain this change?

NanakoAC (changes requested):
some things are questionable from a technical standpoint, the brain healing changes are probably not a good way to achieve that goal

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
JakeDaBoss: Speeding up heal recovery. It's rather slow right now.

code/modules/reagents/Chemistry-Recipes.dm
JakeDaBoss: As per the Changelog:
  - rscadd: "Chemical Reagent: Phlogiston, which upon mixing, unleashes a spreading and debilitating Phoron fire. Made with 1u Aluminum, Phoron and Sulphuric Acid, results in 1u."
  - rscadd: "Chemical Reagent: Vecuronium Bromide, one of the most powerful paralytics. It will reduce someone's voice to whispers and will near-instantly paralyze them. Made with 1u Ethanol, and 2u of Mercury and Hydrazine. Results in 2u."


These are from the newer Bay fork that I have, and were to compensate for the axe that the acids got, as it hamstrung medical/sci-based unitologists in what weaponry was readily available. Besides, this adds diversity.

code/modules/organs/internal/brain.dm
NanakoAC: the brain is supposed to accrue damage relatively fast so that people aren't lying in crit forever. Though we should have the means to save people if they can be recovered

Regarding alkysine, it would be best to rework its code to not be dependant on this at all, just have it fetch the brain from affect_blood

revert these brain healing changes. But if the goal is to make critical injuries more treatable, consider buffing inaprovaline instead. IIRC it's handled in another proc, but it contributes a large quantity to brain oxygenation

JakeDaBoss: I'll have to do some research, but I'll get back to you on this.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Toxins.dm
JakeDaBoss: I'm not sure what you mean with this. The modifier is what technically handles it, based on the threshold. This is why editing the individual values is a saner, safer way to do so than editing the threshold which is a mastervalue for all of them.

These are minor edits.

JakeDaBoss:
@WestHybrid need your sign-off on this.

ghost (approved):

PR #928, opened 2020-10-16
Necromorph Healing and Health Tweaks

NanakoAC:
A package of changes concerning necromorph health.
Firstly, a new mechanic. Some of the incoming damage to necromorphs is now dealt as Lasting Damage, which does not heal and effectively reduces their maximum health. This should help to prevent a necromorph from camping on corruption forever healing off damage and never dying. This applies to all of them except the ubermorph

Secondly, added two new Healing spells which signals can use on necromorphs. 
They both require the target to be standing on corruption, and immobilize them for a few seconds while the healing happens
They both come with their own special drawbacks, in addition to relatively high energy cost and longish cooldowns:

Reconstitute: Useable by signals.  Heals 40 normal damage. Regrows one missing limb, but deals Lasting Damage to the target based on the health of the missing limb. 

Rebuild: Useable only by the marker, and currently the only way to heal lasting damage in the whole game.
Heals all lasting damage, and regrows all limbs, but it invests extra biomass into the necromorph based on the quantities healed. This is regained over time after it dies like normal, biomass is never truly lost.

Lastly, buffs the max health of a few common necros to compensate for the first change

changes: 
  - rscadd: "Necromorphs now take 10% of incoming damage as Lasting Damage, which does not heal and effectively reduces their maximum health."
  - tweak: "Slightly buffed the max health of divider, spitter, slasher, leaper and lurker, to partially compensate for the above."
  - rscadd: "Added two new necromorph healing spells - one for signal, and one for marker. Both come with their own drawbacks, but can regrow lost limbs on live necromorphs."

### DIFF
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -2186,6 +2186,7 @@
 #include "code\modules\mob\observer\freelook\marker\signal_abilities\blowout.dm"
 #include "code\modules\mob\observer\freelook\marker\signal_abilities\breach.dm"
 #include "code\modules\mob\observer\freelook\marker\signal_abilities\flicker.dm"
+#include "code\modules\mob\observer\freelook\marker\signal_abilities\heal.dm"
 #include "code\modules\mob\observer\freelook\marker\signal_abilities\lockdown.dm"
 #include "code\modules\mob\observer\freelook\marker\signal_abilities\placement_ability.dm"
 #include "code\modules\mob\observer\freelook\marker\signal_abilities\psi_energy.dm"
--- a/code/_onclick/hud/healthbar.dm
+++ b/code/_onclick/hud/healthbar.dm
@@ -79,6 +79,7 @@
 		delta_meter = new(src)
 		limit_meter = new(src)
 		textholder = new(src)
+		update(TRUE)
 
 
 /obj/screen/healthbar/proc/set_mob(var/mob/living/newmob)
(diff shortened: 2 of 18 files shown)


Ketrai:
Interesting balance... but this will heavily hamper tier 2+ necromorphs. 

1. Once you're a valuable necromorph, the pressure to keep that necro alive and contribute as long as possible becomes very real. A good tripod is distinguished from a bad one when it comes to restraint. Not engaging bad matchups, or poking back and forth to waste ammo is often better than a frontal assault and a swift death.

2. this will make taking damage feel like you're being punished twice.  You loose health, AND you immediately loose max HP too. This is especially frustrating when facing weapons that can hardly be avoided. (Seeker rifle, line gun, plasma cutter).

3. Puker damage nerf isn't documented.

4. 1kg per 2 HP is a lot. You're only going to want to use heal in combat, to completely restore a tier 3 or tier 2 necromorph. Given the long cooldown and energy cost. Restoring an almost dead tripod can easily incur a 100 KG penalty. That's a bit much. Also, what if the marker has 50 biomass but the cost is 100? Will biomass go negative?

5. Reconstitute has the ability to be rather overpowered on the brute in particular. Imagine an almost dead brute, suddenly coming back to full strength in the middle of  a fight. Tier three necros are the only ones who can break stalemates in some cases.

6. Regular heal is overpowered for some, but weak for others. Who cares about stitching a slasher back together if they'll be even squishier.

Suggestions

- Limit the lasting damage to tier 1 necromorphs. Remove the health buff to tier 2 necromorphs. Prevent healing of T3 necromorphs.

RuneQuester:
Ehhhhh. .  i dont like it from a first read
i guess i'll see in action

AshTheGuy:
Having necros regrow limbs seems kinda silly, but hey! That's just me.

PR #950, opened 2020-10-22
Blast doors now damage and weaken mobs

aguyiguess:
makes blast doors damage and weaken mobs if they're under it when it closes. they're meant to be pretty strong I think this makes sense

### DIFF
--- a/code/game/machinery/doors/blast_door.dm
+++ b/code/game/machinery/doors/blast_door.dm
@@ -78,6 +78,7 @@
 // Parameters: None
 // Description: Opens the door. No checks are done inside this proc.
 /obj/machinery/door/blast/proc/force_open()
+	sleep(10)
 	src.operating = 1
 	playsound(src.loc, open_sound, 100, 1)
 	flick(icon_state_opening, src)
@@ -93,6 +94,7 @@
 // Parameters: None
 // Description: Closes the door. No checks are done inside this proc.
 /obj/machinery/door/blast/proc/force_close()
+	sleep(10)
 	src.operating = 1
 	playsound(src.loc, close_sound, 100, 1)
 	src.layer = closed_layer
@@ -104,8 +106,12 @@
 	sleep(15)
 	src.operating = 0
 	for(var/mob/living/M in src.loc)
-		M.take_organ_damage(20)
-		M.Weaken(3)
+		if(istype(src, /obj/machinery/door/blast/shutters)) //shutters are weaker blast doors so they wouldn't do as much damage
+			M.take_organ_damage(5)
+			M.Weaken(1)
+		else
+			M.take_organ_damage(15)
+			M.Weaken(3)
 
 // Proc: force_toggle()
 // Parameters: None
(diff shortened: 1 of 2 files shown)


Ketrai:
No, this will just make people close blast doors repeatedly on mobs. The only way I can see this being fair is when blast controls have a delay, or if blast doors were out of player control (like to close a window breach)

aguyiguess:
You can easily walk through as it’s closing but I can add a delay to the button

code/game/machinery/doors/blast_door.dm
```diff
@@ -103,6 +105,14 @@
 	src.set_opacity(1)
 	sleep(15)
 	src.operating = 0
+	if(istype(src, /obj/machinery/door/blast/regular)) //regular blast doors are heavier and do more damage
+		for(var/mob/living/M in src.loc)
+			M.take_organ_damage(20)
```
JakeDaBoss, line 110: Please, do not do this. This will cause a LOT of issues, and assured death of anyone unlucky enough.

JakeDaBoss (changes requested):

NanakoAC (changes requested):
This isn't necessarily a bad idea, but it is potentially fraught with balance issues which need to be addressed. Here's how this ought to be done:
1. After closing on a mob, the door needs a delay before it can be opened again. This needs to be longer than the stun time to prevent infinite stunlocking
2. Attempting to close the door on a large mob (brute, tripod, ubermorph) should fail. check mob size and fail if its above normal. The attempt to do so, can still damage and stun them a bit, assuming the same delay is added on the door
3. When applying the damage, use take_overall_damage

Do you reckon this is within the realm of your coding abilities?

code/game/machinery/doors/blast_door.dm
```diff
@@ -78,6 +78,7 @@
 // Parameters: None
 // Description: Opens the door. No checks are done inside this proc.
 /obj/machinery/door/blast/proc/force_open()
+	sleep(10)
```
NanakoAC, line 81: delay should come after the door closes on a mob,  but before setting the door's operating var to false, and needs to be longer

code/game/machinery/doors/blast_door.dm
NanakoAC: you're mistaken here, take organ damage is a generic proc for damaging limbs. However, it is intended only for use when targeting a specific limb, and isn't the right thing here.

use take_overall_damage or apply_damage

code/game/machinery/doors/blast_door.dm
```diff
@@ -103,6 +105,14 @@
 	src.set_opacity(1)
 	sleep(15)
 	src.operating = 0
+	if(istype(src, /obj/machinery/door/blast/regular)) //regular blast doors are heavier and do more damage
+		for(var/mob/living/M in src.loc)
+			M.take_organ_damage(20)
+			M.Weaken(4)
+	else //shutters are lighter and do less damage
+		for(var/mob/living/M in src.loc)
+			M.take_organ_damage(7)
```
NanakoAC, line 114: this damage value can be a lot higher, 30 seems reasonable

NanakoAC (changes requested):

NanakoAC:
requested changes not made, we can't use this as is

PR #956, opened 2020-11-03
Executive shuttle can now only be launched during EVAC procedures.

Kmc2000:
* Executive shuttle can no longer launch whenever it wants, it has to be done during / after evac.


JakeDaBoss (approved):
Thanks for the quick work!

PR #963, opened 2020-11-08
Bugfixing.

JakeDaBoss:
🆑 
  - rscadd: "Added a new proc to message moderators specifically. Not useful for players, but useful for backend tinkering."
  - bugfix: "Fixed communications not being present on Aegis."
  - bugfix: "Fixed SolNet being unreachable on Aegis or the evacuation site."
  - bugfix: "Fixed Rejuvinate not being under the rejuvinate perm. Don't ask, can't explain."
  - tweak: "Added priority announcement by staff under the Rejuvinate permissionsgroup."
  - tweak: "Significantly reduced cargo train speed. No more combat train."
  - tweak: "Duplicate IP and CID accounts are now logged to moderators and admins alike."
  - tweak: "Moved: Delay, Restart and the toggle OOC/LOOC verbs to Senior Moderator permissions."
  - tweak: "MoveD: Activate Marker verb to full moderators, instead of trial moderator."
  - rscdel: "Removed random teleportation after hitting edge of admin or Aegis levels."
🆑

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -10,6 +10,12 @@ var/global/floorIsLava = 0
 	for(var/client/C in GLOB.admins)
 		if(R_ADMIN & C.holder.rights)
 			to_chat(C, msg)
+/proc/message_mods(var/msg)
+	msg = "<span class=\"log_message\"><span class=\"prefix\">MOD LOG:</span> <span class=\"message\">[msg]</span></span>"
+	log_adminwarn(msg)
+	for(var/client/C in GLOB.admins)
+		if(R_MOD & C.holder.rights)
+			to_chat(C, msg)
 /proc/message_staff(var/msg)
 	msg = "<span class=\"log_message\"><span class=\"prefix\">STAFF LOG:</span> <span class=\"message\">[msg]</span></span>"
 	log_adminwarn(msg)
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -26,7 +26,6 @@ var/list/admin_verbs_admin = list(
 	/datum/admins/proc/force_antag_latespawn, //Force a specific template to try a latespawn proc,
 	/datum/admins/proc/toggleenter,		//toggles whether people can join the current game,
 	/datum/admins/proc/toggleguests,	//toggles whether guests can join the current game,
-	/datum/admins/proc/announce,		//priority announce something to all clients.,
 	/client/proc/colorooc,				//allows us to set a custom colour for everythign we say in ooc,
 	/client/proc/admin_ghost,			//allows us to ghost/reenter body at will,
 	/client/proc/toggle_view_range,		//changes how far we can see,
@@ -78,7 +77,6 @@ var/list/admin_verbs_admin = list(
 	/client/proc/free_slot_submap,
 	/client/proc/free_slot_crew,			//frees slot for chosen job,
 	/client/proc/cmd_admin_change_custom_event,
-	/client/proc/cmd_admin_rejuvenate,
 	/client/proc/toggleghostwriters,
 	/client/proc/toggledrones,
 	/datum/admins/proc/show_skills,
@@ -233,7 +231,9 @@ var/list/admin_verbs_permissions = list(
 	/client/proc/drop_bomb
 	)
 var/list/admin_verbs_rejuv = list(
-	/client/proc/respawn_character
+	/client/proc/respawn_character,
+	/client/proc/cmd_admin_rejuvenate,
+	/datum/admins/proc/announce		//priority announce something to all clients.,
 	)
 
 //verbs which can be hidden - needs work
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -26,6 +26,8 @@ var/list/admin_verbs_admin = list(
 	/datum/admins/proc/force_antag_latespawn, //Force a specific template to try a latespawn proc,
 	/datum/admins/proc/toggleenter,		//toggles whether people can join the current game,
 	/datum/admins/proc/toggleguests,	//toggles whether guests can join the current game,
+	/client/proc/cmd_admin_rejuvenate,
+	/datum/admins/proc/announce,		//priority announce something to all clients.,
 	/client/proc/colorooc,				//allows us to set a custom colour for everythign we say in ooc,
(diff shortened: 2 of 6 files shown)


code/modules/admin/admin_verbs.dm
```diff
@@ -233,7 +231,9 @@ var/list/admin_verbs_permissions = list(
 	/client/proc/drop_bomb
 	)
 var/list/admin_verbs_rejuv = list(
-	/client/proc/respawn_character
+	/client/proc/respawn_character,
+	/client/proc/cmd_admin_rejuvenate,
+	/datum/admins/proc/announce		//priority announce something to all clients.,
```
NanakoAC: the announcement verb has nothing to do with rejuvenating, why is it here?

Announce should be available to all staff of every role, not under a special category like this

code/modules/vehicles/cargo_train.dm
```diff
@@ -364,7 +364,7 @@
 		move_delay = max(0, (-car_limit * active_engines) + train_length - active_engines)	//limits base overweight so you cant overspeed trains
 		move_delay *= (1 / max(1, active_engines)) * 2 										//overweight penalty (scaled by the number of engines)
 		move_delay += config.run_speed 														//base reference speed
-		move_delay *= 1.1																	//makes cargo trains 10% slower than running when not overweight
```
NanakoAC, line 367: I agree that the cargo train is a bit overpowered, but this is punitive, and would completely destroy its suitability for the intended role.

it shouldn't be able to run easily over corruption, that'll take some more precise coding to do. But even as a stopgap, this is far too severe

maps/DeadSpace/ishimura.dm
```diff
@@ -6,12 +6,13 @@
 	name = "Ishimura"
 	full_name = "USG Ishimura"
 	path = "ishimura"
-	station_levels = list(1,2,3) // Add 4 to this once the EVA Deck is complete, and remove this comment.
-	contact_levels = list(1,2) // Add 3 to this once the EVA Deck is complete, and remove this comment.
+	station_levels = list(1,2) // Add 3 to this once the EVA Deck is complete, and remove this comment.
+	contact_levels = list(1,2,3,4) // Add 4 to this once the EVA Deck is complete, and remove this comment.
```
NanakoAC: this line fixes comms and solnet, right?

maps/DeadSpace/ishimura.dm
```diff
@@ -6,12 +6,13 @@
...
 	player_levels = list(1,2) // Add 3 to this once the EVA Deck is ocmplete, and remove this comment.
 	admin_levels = list(4) // Change this to 5 once the EVA Deck is complete.
+	sealed_levels = list(4) // No edge-of-map teleportation for players who hit the edge, to prevent shitlery.
```
NanakoAC: zlevel 4 is aegis VII ?

NanakoAC (changes requested):

code/modules/admin/admin_verbs.dm
JakeDaBoss: Not according to Aetsuki.

Aetsuki agreed to have it available to Senior Mods. The only permission that sets Senior Moderators apart that I know of is Rejuvinate. Is it a shoddy way of ensuring? Yes. But at the same time, I can't throw this entire system into disarray for one verb.

code/modules/vehicles/cargo_train.dm
JakeDaBoss: I've altered these values so far. 3.2 is a sane value. Tested it out myself.

maps/DeadSpace/ishimura.dm
JakeDaBoss: Correct.

maps/DeadSpace/ishimura.dm
JakeDaBoss: zlevel 4 is admin / evac site. So transit and evac.

NanakoAC: the aegis VII level should be added to this

JakeDaBoss: Oh, you're right.

NanakoAC (approved):

code/modules/admin/admin_verbs.dm
```diff
@@ -233,7 +234,16 @@ var/list/admin_verbs_permissions = list(
...
+	/client/proc/respawn_character,
+	/client/proc/cmd_admin_rejuvenate,
+	/datum/admins/proc/announce,		//priority announce something to all clients.,
+	/datum/admins/proc/restart,
+	/datum/admins/proc/delay,
+	/datum/admins/proc/toggleooc,		//toggles ooc on/off for everyone,
+	/datum/admins/proc/toggleaooc,		//toggles aooc on/off for everyone,
+	/datum/admins/proc/togglelooc,		//toggles looc on/off for everyone,
+	/datum/admins/proc/toggleoocdead,	//toggles ooc on/off for everyone who is dead,
+	/datum/admins/proc/toggledsay		//toggles dsay on/off for everyone,
```
NanakoAC, line 246: these various communicate procs don't have much to do with rejuvenating, why not just give them to moderators instead?

code/modules/admin/admin.dm
```diff
@@ -10,6 +10,12 @@ var/global/floorIsLava = 0
 	for(var/client/C in GLOB.admins)
 		if(R_ADMIN & C.holder.rights)
 			to_chat(C, msg)
+/proc/message_mods(var/msg)
```
NanakoAC, line 13: Shouldn't this message people of rank mod -or higher- ? 
I can't see any reason you'd want to specifically message a moderator but not people with higher permissions. Admins, hosts, developers, etc.

As far as i can tell, the usage of this is going to limit the propagation of messages

code/modules/admin/admin_verbs.dm
JakeDaBoss: +MOD or R_MOD in this case is handed out to regular moderators, including trial moderators. This current change puts it squarely under Senior Mod permissions. Prior, admins had this permission as only one. Rather than a three-rank downward bump on permissions, I just made it singular. Better odds of getting it past the headmins.

code/modules/admin/admin.dm
JakeDaBoss: This should inherit. R_MOD is something that admins have.

NanakoAC (approved):

PR #977, opened 2020-11-16
Bugfixing

JakeDaBoss:
🆑 
  - tweak: "Changed the Lurker's mob size status to MOB_SMALL as opposed to MOB_MEDIUM."
  - tweak: "Altered locker code. It will not fit one (1) medium sized mob."
🆑


JakeDaBoss:
Related in this update:

There's a particular issue with certain Necromorphs (currently being tested & confirmed with Slashers) being allowed to dogpile into a locker using "Rest". This has so far been confirmed with three Necromorphs in a singular locker.

Locker code runs on mob_size. This is why I now universally applied this to all Necromorphs.

code/modules/mob/living/carbon/human/species/necromorph/exploder.dm
```diff
@@ -12,6 +12,8 @@
 	total_health = 85	//It has high health for the sake of making it a bit harder to destroy without targeting the pustule. Exploding the pustule is always an instakill
 	biomass = 70
 	mass = 50
+	mob_size = MOB_LARGE
```
NanakoAC: Large size category is only for tier 3+, generally.

Most necromorps should be medium, including exploder, puker, spitter, twitcher, etc.

Large is for brute, tripod, ubermorph and (in future) hunter + pregnant

NanakoAC: Farther to that, this already exists in species.dm
var/mob_size	= MOB_MEDIUM

so medium is the default size, theres no need to specify it, only to specify other values.
remove all size changes except lurker

NanakoAC (changes requested):
size category issues

code/modules/mob/living/carbon/human/species/necromorph/exploder.dm
JakeDaBoss: Got it. But then we need to look into Locker code.

	var/storage_capacity = 2 * MOB_MEDIUM //This is so that someone can't pack hundreds of items in a locker/crate

I'd like to change this, but I'm not sure what 2 * MOB_MEDIUM is.

NanakoAC: that should be enough to fit two medium sized mobs, which is a bit silly. 

try changing it to 1.5

NanakoAC (approved):

PR #1011, opened 2020-11-22
Ubermorph, Linecutter, BPL, balance and bugfixing

### LINKED ISSUE #936: [BUG] Ubermorph abilities can break
#### Description of issue:
As ubermorph, having the ability menu open causes you to inherit verbs of other necros when they're nearby, and causes your own to break.

#### Steps to reproduce:
Be uber, have other necros near, try to use abilities. 
- sense seems to work, same for battle cry when this happens.
- Lunge seems to break after one use.

NanakoAC:
changes: 
  - tweak: "Buffed the bioprosthetic lab modestly all around. Organs grow slightly faster, the growth tank holds more biomass, recycling tank breaks down organic matter more quickly, and grown organs require less to maintain."
  - tweak: "Significantly reduced the flight speed of line cutter projectiles, making them easier to dodge. Damage slightly reduced"
  - bugfix: "Fixed the ubermorph's lunge, regen and sense abilities all being broken"

Closes #936

### DIFF
--- a/code/game/machinery/ds13/bpl/growth_tank.dm
+++ b/code/game/machinery/ds13/bpl/growth_tank.dm
@@ -10,7 +10,7 @@
 	icon = 'icons/obj/machines/ds13/bpl.dmi'
 	icon_state = "base"
 
-	var/max_biomass = 100	//1 Litre
+	var/max_biomass = 150	//1.5 Litre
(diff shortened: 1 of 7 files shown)


Ketrai:
😃

PR #1015, opened 2020-11-24
Sound Environment work

JakeDaBoss:
This update aims to add onto existing systems by differentiating room sound environments and how sounds reflect and are perceived inside these spaces.

### DIFF
--- a/maps/DeadSpace/Aegis_areas.dm
+++ b/maps/DeadSpace/Aegis_areas.dm
@@ -4,41 +4,53 @@
 	name = "\improper Aegis VII"
 	icon = 'maps/DeadSpace/ishimura_areas.dmi'
 	icon_state = "aegis"
+	sound_env = ASTEROID
 
 /area/aegis/colony
 	name = "\improper Aegis Colony"
 	icon_state = "crewbase"
+	sound_env = LARGE_ENCLOSED
 
 /area/aegis/colony/airlock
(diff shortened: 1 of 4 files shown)


maps/DeadSpace/Aegis_areas.dm
```diff
@@ -4,41 +4,53 @@
 	name = "\improper Aegis VII"
 	icon = 'maps/DeadSpace/ishimura_areas.dmi'
 	icon_state = "aegis"
+	sound_env = ASTEROID
```
NanakoAC, line 7: While we've not gotten around to fixing this yed, aegis VII is not an asteroid. It's not meant to be the lightless, airless rock it currently is. Aegis VII in canon is a rocky desert planet with a breathable atmosphere and a harsh noonday sun

NanakoAC (commented):
small comment about aegis

maps/DeadSpace/Aegis_areas.dm
JakeDaBoss: This is for a sound environment. It's for the reverb and sound reflection to take effect.

This is simply a very, very large open area basically. So, the issue you see is a non-issue.

JakeDaBoss:
Remind me to kick @Snypehunter007's ass for not doing this to begin with, big rees!

PR #1017, opened 2020-11-24
Ambience (Alpha)

theresidentnine:
This is still fairly early and needs some changing, but it is basic test of ambience.

Three background tracks have been added: standard one that plays on the ship, maintenance, and 'freezer' which is for both freezers and cryo. I will add more but we need to know if this is good first.

The button/pager sound was also changed to a Dead Space one.

Big bug is the sounds restarting when entering another area that still has the same ambience track. Could also use more fade out/in.


JakeDaBoss:
Right-on. Did not expect this to cause issues, but just update this branch to the latest dev branch, and merge all changes. Your own, and the dev ones.

Will review now.

code/game/sound.dm
```diff
@@ -51,7 +51,7 @@ GLOBAL_LIST_INIT(lighter_sound,list('sound/items/lighter1.ogg','sound/items/ligh
 GLOBAL_LIST_INIT(keyboard_sound,list('sound/machines/keyboard/keypress1.ogg','sound/machines/keyboard/keypress2.ogg','sound/machines/keyboard/keypress3.ogg','sound/machines/keyboard/keypress4.ogg'))
 GLOBAL_LIST_INIT(keystroke_sound,list('sound/machines/keyboard/keystroke1.ogg','sound/machines/keyboard/keystroke2.ogg','sound/machines/keyboard/keystroke3.ogg','sound/machines/keyboard/keystroke4.ogg'))
 GLOBAL_LIST_INIT(switch_sound,list('sound/machines/switch1.ogg','sound/machines/switch2.ogg','sound/machines/switch3.ogg','sound/machines/switch4.ogg'))
-GLOBAL_LIST_INIT(button_sound,list('sound/machines/button1.ogg','sound/machines/button2.ogg','sound/machines/button3.ogg','sound/machines/button4.ogg'))
```
JakeDaBoss, line 54: Why were these removed? They add variety, and nothing added in their stead.

Shouldn't do this without replacements.

code/game/machinery/pager.dm
```diff
@@ -26,7 +26,7 @@
 /obj/machinery/pager/attack_hand(mob/living/user)
 	if(..()) return 1
 	if(istype(user, /mob/living/carbon))
-		playsound(src, "button", 60)
+		playsound(src, "button", 20)
```
JakeDaBoss, line 29: This volume was fine. No need to nearly mute it.

code/game/machinery/lightswitch.dm
```diff
@@ -61,7 +61,7 @@
 		return 1
 
 /obj/machinery/light_switch/attack_hand(mob/user)
-	playsound(src, "switch", 30)
+	playsound(src, "button", 20)
```
JakeDaBoss, line 64: This volume was fine. No need to nearly mute it.

code/game/machinery/buttons.dm
```diff
@@ -33,7 +33,7 @@
 /obj/machinery/button/attack_hand(mob/living/user)
 	if(..()) return 1
 	if(istype(user, /mob/living/carbon))
-		playsound(src, "button", 60)
+		playsound(src, "button", 20)
```
JakeDaBoss, line 36: This volume was fine. No need to nearly mute it.

maps/DeadSpace/Ishimura_areas.dm
```diff
@@ -2,6 +2,7 @@
 /area/ishimura
 	icon = 'maps/DeadSpace/ishimura_areas.dmi'
 	ship_area = TRUE
+	forced_ambience = list('sound/ambience/dsamb_standardcorridor.ogg') //by default in the ship we play the corridor sound.
```
JakeDaBoss, line 5: Do not use forced_ambience. It will cause looping or jarring issues in relation to ambience changes.

Please simply use ambience =, and not forced_ambience =.

JakeDaBoss (changes requested):

PR #1035, opened 2020-12-01
More Sounds & Fixes

theresidentnine:
soundadd: External doors (the red ones!) as well as high security and vault doors have new Dead Space heavy door sounds.
soundadd: Lightswitches and buttons have new Dead Space beep sounds.
soundadd: All vending machines have new Dead Space browsing and purchase sounds.
soundadd: A variety of Dead Space sounds for biotanks, and one for recycler interactions too.
bugfix: Fixed bad punctuation in recycling tank interactions.
bugfix: Removed some unused duplicate .oggs from the files, every little helps.


theresidentnine:
I'll add video to show this shortly.

theresidentnine:
https://cdn.discordapp.com/attachments/693569828992057384/783454858291183616/Project_1.webm

JakeDaBoss (approved):
Looks good.

theresidentnine:
https://cdn.discordapp.com/attachments/693569828992057384/784146797811466240/Project_1.webm

JakeDaBoss (changes requested):
Changelog missing.

theresidentnine:
> Changelog missing.

I have updated the changelog at top already, it is just this that's new:
soundadd: A variety of Dead Space sounds for biotanks, and one for recycler interactions too.

code/game/machinery/buttons.dm
```diff
@@ -33,7 +33,7 @@
 /obj/machinery/button/attack_hand(mob/living/user)
 	if(..()) return 1
 	if(istype(user, /mob/living/carbon))
-		playsound(src, "button", 60)
```
NanakoAC, line 36: these are intentionally loud so they can be heard from a decent distance away. 
Don't type in specific numbers at all, use the VOLUME_XXX defines found in defines/audio

Buttons should be high since they're a player initiated action which can often do stuff of significant tactical consequence

code/game/machinery/doors/airlock_control.dm
```diff
@@ -269,6 +269,7 @@ obj/machinery/access_button/attackby(obj/item/I as obj, mob/user as mob)
 
 obj/machinery/access_button/attack_hand(mob/user)
 	..()
+	playsound(src, "button", 20)
```
NanakoAC: as above, apply volume defines to everything in this file

code/modules/mob/observer/freelook/marker/marker.dm
```diff
@@ -4,6 +4,7 @@
 */
 /obj/machinery/marker
 	name = "Marker"
+	desc = "A strange and imposingly large relic of unknown origin."
```
NanakoAC: leave this with no description, the marker is meant to be mysterious

NanakoAC (changes requested):
One additional general comment. While removing duplicates is fine, please make sure that only duplicates get removed. Unused assets should be kept for future use

theresidentnine:
I've reverted the marker description but I lowered volume because the replacement Dead Space sounds are VERY LOUD files.

Using the inbuilt sound code controls is the most intuitive way of solving this. I've worked hard to make sure they're around the same volume/range as what they replace.

I don't like the defines because, for example, 30 of volume_mid_high is too loud and 18 of volume_mid is too quiet. Most sound in the code doesn't use those defines anyway.

JakeDaBoss:
> > Changelog missing.
> 
> I have updated the changelog at top already, it is just this that's new:
> soundadd: A variety of Dead Space sounds for biotanks, and one for recycler interactions too.

You're missing a changelog .yml

NanakoAC:
> I've reverted the marker description but I lowered volume because the replacement Dead Space sounds are VERY LOUD files.

Very well then, reducing the volume is fine, but still use defines

> I don't like the defines because, for example, 30 of volume_mid_high is too loud and 18 of volume_mid is too quiet. Most sound in the code doesn't use those defines anyway.

Most sound in the code was done before these defines were added. This is a new development policy going forward and every future addition should work with it

 They're the numbers they are for a reason. The volume scale is linear, but audio scales logarithimically. The difference between 1 and 3 volume is far more significant than the difference between 70 - 100. There's a lot of math to it, and these values are a little fuzzy, but they represent volume scaling far more accurately than adjusting numbers manually

You can't pick volume by selecting numbers that seem right, a number that's the average of two values is not actually inbetween them

Pick the define whose name roughly represents how loud you want the sound to be, and don't think about the numbers at all

If you're interested in an academic dissection of the topic, this is a good introduction point
https://www.dr-lex.be/info-stuff/volumecontrols.html

theresidentnine:
Done

NanakoAC (approved):

PR #1037, opened 2020-12-02
Seeindark bugfix

### LINKED ISSUE #1027: [BUG] Necros without any degree of night-vision get dunked in darkness.
#### Description of issue:
It's pitch black.

To be able to see just a little, please?

#### Steps to reproduce:
Be slasher.
Become maint dweller.
Throw a temper tantrum, smash some lights.
End up being an involuntary void dweller.

JeserTheFox:
bugfix: "Removed see_in_dark value changes that resulted in necromorphs not having dark vision."

Removed any found see_in_dark changes and set it to 255 for mob/living.

Closes #1027

### DIFF
--- a/html/changelogs/seeindark_bugfix.yml
+++ b/html/changelogs/seeindark_bugfix.yml
@@ -1,0 +1,6 @@
+author: Jeser
+
+delete-after: True
+
+changes: 
+  - bugfix: "Removed see_in_dark value changes that resulted in necromorphs not having dark vision."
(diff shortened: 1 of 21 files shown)


html/changelogs/seeindark_bugfix.yml
```diff
@@ -0,0 +1,6 @@
+author: Jeser
+
+delete-after: True
+
+changes: 
+  - bugfix: "Removed see_in_dark value changes that resulted in necromorphs not having dark vision."
```
NanakoAC, line 6: Very minor comment, i wont require this change but:

Generally when writing changelogs, remember that they're going to be read by users with no technical knowledge. In this case it should say something like "Fixed the bug that caused necromorph nightvision to not work in complete darkness"

NanakoAC (approved):

PR #1066, opened 2021-01-02
Pop-up text on mob.

DeadLineSS13:
Add pop-up text on mob.
unknown


NanakoAC (changes requested):
Alright, after testing it out, i've noted a few small problems:
1. It's not enabled by default, most people will never notice it without that, turn it on.
2. The line spacing between the text is a bit much, easy to make big towers, try reducing that
3. The time that it stays onscreen is too short, barely time to read it. Increase by 50%
4. Radio messages are shown there too. Generally, they should only be shown to people who were close enough to hear the message

JakeDaBoss:
Besides what Nanako put up, did you confirm that Humans cannot see Necro's pop-ups, or does it happen for them at all?

DeadLineSS13:
> Besides what Nanako put up, did you confirm that Humans cannot see Necro's pop-ups, or does it happen for them at all?

Nanako checked this.

code/modules/mob/living/say.dm
```diff
@@ -264,15 +264,16 @@ proc/get_radio_key_from_channel(var/channel)
 	var/list/listening_clients = list()
 	for(var/mob/M in listening)
 		if(M)
-			M.hear_say(message, verb, speaking, alt_name, italics, src, speech_sound, sound_vol)
-			if(M.client)
-				listening_clients += M.client
+			if(M.hear_say(message, verb, speaking, alt_name, italics, src, speech_sound, sound_vol))
+				if(M.client)
```
NanakoAC: hear_say also includes a check for client so this one is unnecessary

NanakoAC (approved):

PR #1086, opened 2021-01-05
Tweak for shooting at most of not density items

Terror4000rus:
- tweak: "If a gun is firing at catwalks or most of items (not structures), its projectiles will not count it as target (no more firing at catwalks)."


code/modules/projectiles/gun.dm
```diff
@@ -327,6 +327,8 @@
 
 
 
+	if(istype(target, /obj/structure/catwalk) || istype(target, /obj/item))
```
NanakoAC: Putting specific typechecks into unrelated code like this is bad. This generates unmaintainable spaghetti code

Find a better approach

NanakoAC (changes requested):
Not a good implementation, this would be better done with an atom flag to mark an atom as untargetable

code/modules/projectiles/gun.dm
Terror4000rus: Done

code/game/objects/items.dm
```diff
@@ -99,6 +99,8 @@
 	can_block_movement = FALSE
 
 /obj/item/New()
+	atom_flags |= ATOM_FLAG_UNTARGETABLE //items on floor shouldn't be a target for guns
```
NanakoAC: Flags are set at authortime, not in New
Set them in the block above

Figure out which other flags items have by default and append this one

Doing it in New is not good because this code will prevent the possibility of making targetable items, which is very much needed. For example, computers are all items, even the dense anchored consoles. Make sure to set the atom flags on those

code/modules/atmospherics/atmospherics.dm
```diff
@@ -33,6 +33,7 @@ Pipelines + Other Objects -> Pipe network
 	var/atmos_initalized = FALSE
 
 /obj/machinery/atmospherics/New()
+	atom_flags |= ATOM_FLAG_UNTARGETABLE //vents, scrubbers and pipes shouldn't be a target for guns
```
NanakoAC: as above, not in New

NanakoAC (changes requested):
implementation still has some technical flaws

code/game/objects/items.dm
Terror4000rus: > Figure out which other flags items have by default and append this one

Sorry, but I don't have that much time to check every item in the game for atom_flags.

> Doing it in New is not good because this code will prevent the possibility of making targetable items, which is very much needed. For example, computers are all items, even the dense anchored consoles. Make sure to set the atom flags on those

It can be done just by this:
```
/obj/item/New()
   . = ..()
  atom_flags &= ~ATOM_FLAG_UNTARGETABL
```
Some items override atom_flags, which will cause deleting of the flag. So I did it to ensure, that every item in subtype will have it (unless removed with a new New).
Implecation with the flag just in main object:
xYwn8981Vi
Vials override it.

code/modules/atmospherics/atmospherics.dm
Terror4000rus: As above.

Terror4000rus:
I gave comments

code/game/objects/items.dm
NanakoAC: this is not how things are done, having code in a base class to set or unset things gets messy, it creates traps for the unaware coder, and it will be hell to track down in future. It is generally a violation of coding standards.

You dont have to check every item in the game, just the base item class and set it there. If various things are setting their own flags and wouldn't be affected by this change, so be it. those problems can be dealt with in future, when and if they arise.

Trying to pre-emptively solve them with this approach is going to create worse problems in the long run

NanakoAC:
comments responded to, the approach still needs tweaking

Terror4000rus:
Alright. I just reverted for everything but catwalks. @NanakoAC

NanakoAC (approved):

PR #1089, opened 2021-01-05
Banning writing in blood and drawing runes in line of sight

Terror4000rus:
- rscadd: "Signals can't write in blood and draw runes in line of human sight."

Why?
It looks strange and frustrating in game. You are trying to play your role while some signal decides to write under your feets "DIE DIE DIE" or 2-3 runes at walls and floor.
...And just unfitting. It makes more sense for a theory that a mad crewmember wrote them than the marke~~t~~r's invisible hand.

### DIFF
--- a/code/modules/mob/observer/freelook/marker/signal_abilities/blood_writing.dm
+++ b/code/modules/mob/observer/freelook/marker/signal_abilities/blood_writing.dm
@@ -6,7 +6,7 @@
 	energy_cost = 15
 	require_corruption = FALSE
 	autotarget_range = 0
-	LOS_block = FALSE	//This is for spooking people, we want them to see it happen
+	LOS_block = TRUE
 	target_types = list(/turf/simulated)
 
 
--- a/code/modules/mob/observer/freelook/marker/signal_abilities/blood_writing.dm
+++ b/code/modules/mob/observer/freelook/marker/signal_abilities/blood_writing.dm
@@ -19,14 +19,26 @@
 		return
 	return TRUE
 
+/datum/signal_ability/writing/proc/middle_check(var/mob/user, var/turf/target)
+	if(LOS_block) //copied from is_valid_target
+		var/mob/M = target.is_seen_by_crew()
+		if (M)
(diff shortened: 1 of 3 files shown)


NanakoAC (changes requested):
It's equally frustrating to try to place things around people who are camping doorways to maximise their vision radius, this is why i changed runes to spawn in when unseen.

If you want to change writing, make them use that same system, though it'll be a bit more complex since a message is needed

The current approach at least ensures messages will be read, this proposed change would decrease engagement

Terror4000rus:
> this is why i changed runes to spawn in when unseen.

what
I literally changed them to make their appearance unseen in line of sight now.
I don't understand what you want to see here

NanakoAC:
look at how the rune spell works

Terror4000rus:
> look at how the rune spell works

I still don't get it, but I'll do a thing and ping you, maybe it will make it smother

Terror4000rus:
@NanakoAC

NanakoAC:
My apologies, I got mixed up with some code in the sanity branch which is not merged yet, i thought i'd done it in a seperate PR last month. The thing i said to go look at with runes is not actually in dev yet, so i sent you on a wild goose chase by mistake, i'll try to avoid this kind of confusion in future

So changes to how runes spawn will come with the sanity project, and blood writing can be changed after that. 
In case you're interested, the planned system in future is that if a signal attempts to spawn these things when someone is looking, it instead places an invisible marker, which will automatically finish the spawn when nobody is around

I stick by my second statment though, this PR as is will reduce engagement with the feature, and shouldn't be implemented even as a temporary stopgap

Terror4000rus:
Alright, with such changes in future, I'm satisfied, but

> I stick by my second statment though, this PR as is will reduce engagement with the feature, and shouldn't be implemented even as a temporary stopgap

Not every item or game mechanic have to be in game if it is actually harmful for a player's experience.
Some things have to be removed temporary, especially if devs don't have time and motivation to make them better.

PR #1181, opened 2021-01-12
Executive shuttle shutters button fix + various window replacements

KeRSedChaplain:
Makes the executive shuttle shutters button function properly to open/close the shutters in exec escape
Replaces various old style windows with full tile windows


JakeDaBoss (approved):
As these fixes were asked by me, I don't see an issue in possible merge conflicts. Well done, and thank you.

PR #1182, opened 2021-01-12
Resolves some re-occuring runtimes

Kmc2000:
Seems to just be dead code, and it causes errors, feel free to correct me if I'm wrong.

### DIFF
--- a/code/datums/outfits/outfit.dm
+++ b/code/datums/outfits/outfit.dm
@@ -272,11 +272,14 @@ var/list/outfits_decls_by_type_
 
 //Wrapper for creating, so that we can manipulate the items
 /decl/hierarchy/outfit/proc/create_item(var/path, var/location, var/dummy = FALSE)
+	/* As far as I can tell, this is dead code and just causes runtimes.
 	var/datum/thing = path
+
 	if (dummy && initial(thing.implements_dummy))
 		return new path(location, dummy = TRUE)
 	else
-		return new path(location)
+	*/
+	return new path(location)
 
 
 /decl/hierarchy/outfit/proc/copy()
--- a/code/datums/outfits/outfit.dm
+++ b/code/datums/outfits/outfit.dm
@@ -272,14 +272,12 @@ var/list/outfits_decls_by_type_
 
 //Wrapper for creating, so that we can manipulate the items
 /decl/hierarchy/outfit/proc/create_item(var/path, var/location, var/dummy = FALSE)
-	/* As far as I can tell, this is dead code and just causes runtimes.
 	var/datum/thing = path
 
 	if (dummy && initial(thing.implements_dummy))
 		return new path(location, dummy = TRUE)
 	else
-	*/
-	return new path(location)
+		return new path(location)
 
 
 /decl/hierarchy/outfit/proc/copy()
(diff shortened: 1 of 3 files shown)


code/datums/outfits/outfit.dm
```diff
@@ -272,11 +272,14 @@ var/list/outfits_decls_by_type_
 
 //Wrapper for creating, so that we can manipulate the items
 /decl/hierarchy/outfit/proc/create_item(var/path, var/location, var/dummy = FALSE)
+	/* As far as I can tell, this is dead code and just causes runtimes.
 	var/datum/thing = path
```
NanakoAC, line 275: no, this is newly added code, and very much active. Fix the bugs with it, don't remove it

code/modules/mob/observer/freelook/marker/signal_abilities/signal_ability.dm
```diff
@@ -147,11 +147,10 @@
 //Path to the end of the cast
 /datum/signal_ability/proc/finish_casting(var/mob/user, var/atom/target,  var/list/data)
 	//Pay the energy costs
-	if (!pay_cost(user))
+	if (!user || !pay_cost(user))
```
NanakoAC: if you're going to tweak this, properly abort the casting and refund the cost

code/modules/mob/observer/freelook/marker/signal_abilities/signal_ability.dm
```diff
@@ -162,7 +161,7 @@
 /datum/signal_ability/proc/stop_casting(var/mob/user)
 
 	//Search the user's clickhandlers for any which have an id matching our type, indicating we put them there. And remove those
-	for (var/datum/click_handler/CH in user.GetClickHandlers())
+	for (var/datum/click_handler/CH in user?.GetClickHandlers())
```
NanakoAC, line 163: what does this syntax do? I've never seen it used in this manner

NanakoAC (changes requested):
few changes

code/datums/outfits/outfit.dm
Kmc2000: Thing is, I can't see anywhere where it's used, hence the "bag argument" for dummy. 

Removing it also seemed to not do anything? Weird.

code/modules/mob/observer/freelook/marker/signal_abilities/signal_ability.dm
Kmc2000: Alright

code/modules/mob/observer/freelook/marker/signal_abilities/signal_ability.dm
Kmc2000: It's a left hand null coalescing operator.

if(user && user.GetClickHandlers()) can be simplified to user?.GetClickHandlers()

code/datums/outfits/outfit.dm
NanakoAC: it was newly added for future support to prevent full initialisation of items that are only for visuals

if its causing runtimes, fix those

Kmc2000: It's runtiming because of the new with dummy as a param, but datum doesn't expose a constructor taking in dummy, thus it's causing runtimes inside of BYOND itself. I can extend this to /datum I suppose, but I again am struggling to see what this does, could you point me to a file where it's implemented?

NanakoAC: This is the sister implementation to the code in loadout_gear.dm. See the code in there
I implemented a var, implements_dummy, which indicates if its safe to pass a dummy parameter to that datum at initialisation.

It is currently being used for rigs and PDAs, to prevent them adding themselves to processing lists when created for mannequin previews.

I may not have finalised that half of the implementation, so fixing it would be good.
If you can think of a more elegant way to approach this problem, i'm all for throwing it out and doing it better

NanakoAC (approved):

PR #1184, opened 2021-01-13
Mining RIG addition + minor civ, engi, and advanced suit edits

KeRSedChaplain:
-Adds the intermediate hazard suit, or the mining suit. As of this moment it's not mapped to spawn anywhere yet, but it does work properly.

-Also makes the civillian rig have less acid protection since it was using the base amount of 2 before, which is the same as the vintage suit's acid defense. The advanced suit now has 95 radiation protection instead of the normal engi suits, which now have 75 radiation protection.


code/modules/clothing/spacesuits/rig/suits/engineering.dm
```diff
@@ -2,7 +2,7 @@
 	name = "engineering rig"
 	desc = "A lightweight and flexible armoured rig suit, designed for mining and shipboard engineering."
 	icon_state = "ds_engineering_rig"
-	armor = list(melee = 45, bullet = 60, laser = 60, energy = 25, bomb = 60, bio = 100, rad = 95)
+	armor = list(melee = 45, bullet = 60, laser = 60, energy = 25, bomb = 60, bio = 100, rad = 75)
```
NanakoAC: The engineering rig was accidentally missed in armor rebalancing work, and is too powerful

lower the first three values to 40, 50, 50

code/modules/clothing/spacesuits/rig/suits/mining.dm
```diff
@@ -0,0 +1,36 @@
+/obj/item/weapon/rig/mining
+	name = "intermediate hazard suit"
+	desc = "A stripped down, lighter version of the engineering suit. Commonly used during planet cracking operations or other \
+	 dangerous working environments to keep the wearer safe, while keeping their mobility."
+	icon_state = "ds_mining_rig"
+	armor = list(melee = 35, bullet = 50, laser = 50, energy = 15, bomb = 70, bio = 100, rad = 50)
```
NanakoAC: set melee,  bullet and laser armor to 7.5 points below the engineering suit, after the above changes

code/modules/clothing/spacesuits/rig/suits/engineering.dm
```diff
@@ -2,7 +2,7 @@
...
 	offline_slowdown = 4
 	online_slowdown = 1.75
```
NanakoAC, line 7: set this to 2

code/modules/clothing/spacesuits/rig/suits/mining.dm
```diff
@@ -0,0 +1,36 @@
+/obj/item/weapon/rig/mining
```
NanakoAC, line 1: Add this rig to the loot list in random/rig

also add it to the contents of /obj/structure/closet/secure_closet/planet_cracker
found in code/game/objects/structures/crates_lockers/closets/secure/mining.dm

NanakoAC (commented):
lots of little tweaks, in future we'll have some mapping work to do with this

code/modules/clothing/spacesuits/rig/suits/civilian.dm
```diff
@@ -5,6 +5,7 @@
 	armor = list(melee = 10, bullet = 10, laser = 10, energy = 10, bomb = 10, bio = 10, rad = 10)
 	offline_slowdown = 1
 	online_slowdown = 0
+	acid_resistance = 1
```
NanakoAC: Revert this change; the reason being that this value doesn't work in isolation. It only affects the acid damage that this rig and its pieces take. And since it has no pieces which cover the body, it will absorb relatively little, we just won't want rigs getting easily melted

KeRSedChaplain:
ack, didn't mean to undraft this

ghost (approved):

PR #1196, opened 2021-01-17
Add Agony Scream and sounds for emotes.

DeadLineSS13:
## About PR
This thing creates the feeling that the doll is not a silent entity.


NanakoAC:
To echo what i said in discord:
These scream sounds are not bad, there isn't an issue with their quality. But they dont seem conducive to roleplay, since it means everyone has the same voice. And probably not what they imagined for their character
I'm not opposed to the idea of voice audio for humans, but I think we'd need multiple complete sets of sounds - at least three per gender, plus some frequency customisation, to allow the necessary personalisation for roleplay
the implementation as is, seems a bit lacking in that regard

As long as people lack the ability to customise voice to fit a reasonable approximation of what they'd want their character to sound like, this feature would just be a meme, and actively harmful to RP

NanakoAC:
this feature won't be added in its current state, needs significantly more depth and flexibility as described, so closing for now

PR #1203, opened 2021-01-19
[s] Patches an exploit with signal absorb

Kmc2000:
* Signal absorb is no longer exploitable for unexpected biomass gain

### DIFF
--- a/code/__defines/flags.dm
+++ b/code/__defines/flags.dm
@@ -33,8 +33,6 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define OBJ_FLAG_ANCHORABLE              0x0001 // This object can be stuck in place with a tool
 #define OBJ_FLAG_CONDUCTIBLE             0x0002 // Conducts electricity. (metal etc.)
 #define OBJ_FLAG_INVINCIBLE				 0x0004	// This object ignores all damage mechanics, and can't be broken, damaged etc
-#define OBJ_FLAG_BEING_ABSORBED			 0x0008	// This object is currently being absorbed as biomass.
-
 
 #define MOB_FLAG_HOLY_BAD                0x001  // If this mob is allergic to holiness
 
--- a/code/__defines/flags.dm
+++ b/code/__defines/flags.dm
@@ -33,6 +33,7 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define OBJ_FLAG_ANCHORABLE              0x0001 // This object can be stuck in place with a tool
 #define OBJ_FLAG_CONDUCTIBLE             0x0002 // Conducts electricity. (metal etc.)
 #define OBJ_FLAG_INVINCIBLE				 0x0004	// This object ignores all damage mechanics, and can't be broken, damaged etc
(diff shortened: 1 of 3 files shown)


code/__defines/flags.dm
```diff
@@ -33,6 +33,7 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define OBJ_FLAG_ANCHORABLE              0x0001 // This object can be stuck in place with a tool
 #define OBJ_FLAG_CONDUCTIBLE             0x0002 // Conducts electricity. (metal etc.)
 #define OBJ_FLAG_INVINCIBLE				 0x0004	// This object ignores all damage mechanics, and can't be broken, damaged etc
+#define OBJ_FLAG_BEING_ABSORBED			 0x0008	// This object is currently being absorbed as biomass.
```
NanakoAC: This definitely doesn't need an object flag. this kind of special case is why we have extensions. Just stick a special extension on an object while its being absorbed, and check for the presence of that before absorbing it

code/modules/mob/observer/freelook/marker/biomass.dm
```diff
@@ -11,7 +11,7 @@
 
 /datum/biomass_source/New(var/datum/_source = null, var/datum/_target = null, var/total_mass = 0, var/duration = 1 SECOND)
 	.=..()
-	source = "\ref[_source]"
+	source = "\ref[_source]" //I have no idea what's going on with this, but I'm too scared to change it. Please review this change if you know why it was implemented this way.
```
NanakoAC: this is just a weak pointer, to hold a reference to an object without messing with its GC

NanakoAC (changes requested):
use extensions, flags aren't needed for this

code/__defines/flags.dm
Kmc2000: I felt like a blank extension being used like a flag for this one case is too redundant, a flag like this is super cheap to use and equally easy to check for.

code/modules/mob/observer/freelook/marker/biomass.dm
Kmc2000: ...Weird.

So the biomass source can still continue with a nullpointer if the object is deleted? 

BYOND's \ref is a very weird beast, I'd consider refactoring this to use a straight ref but this sounds intentional.

code/__defines/flags.dm
NanakoAC: It is exactly the opposite. Yes flags are cheap to use, thats why they're reserved for performance critical applications. By adding a new flag there you're reserving it forever and leaving one less potential flag for other, broader applications.

This specific case is too narrow a focus, and its a perfect use case for extensions. The extension gets added and removed a second later, only one thing will ever check it.  Extensions even have list-only forms for this kind of purpose, making them also ultra cheap to add and check for

Don't add new behaviour to everything for the sake of one limited application

code/modules/mob/observer/freelook/marker/biomass.dm
NanakoAC: Well in some cases, it depends on the specific source. Some (like absorbing a human corpse) will delete themselves if they find their source atom is gone. While others (like reclaiming dead necromorphs) will just not do certain behaviours and continue working as normal if their source atom is deleted.

The latter case is to prevent people gibbing corpses (and in the cases of exploders which gib themselves)

NanakoAC (approved):

PR #1204, opened 2021-01-19
Small Storage

Sibg:
Adds New Small Storage Area


NanakoAC (approved):
We'll have to get mapmerge working to implement the tram after this, but that looks feasible now

PR #1206, opened 2021-01-20
Cleans up psychic tracer deletion

### LINKED ISSUE #1040: Necrovision causes long deletion times
ADMIN LOG: Error: /obj/effect/psychic_tracer([0x200c884]) took longer than 1 second to delete (took 2.2 seconds to delete).
Connection failed.
Reconnecting in 10 seconds...

This is almost certainly caused by necrovision updates, need to make those asynchronous and not hold up the caller, using addtimer

Kmc2000:
fixes #1040

* Psychic tracers were forcing a hard reference clear by the GC, which is a lot of work. I've shifted it to free up its memory refs in its destructor

The recursive timer based tick() system is a bit weird, I'd like to monitor that and see if it needs any work. My attempts to get those extensions processing failed for unknown reasons.

### DIFF
--- a/code/modules/mob/observer/freelook/marker/signal_abilities/psychic_tracer.dm
+++ b/code/modules/mob/observer/freelook/marker/signal_abilities/psychic_tracer.dm
@@ -50,7 +50,9 @@
 /obj/effect/psychic_tracer/get_visualnet_tiles(var/datum/visualnet/network)
 	return EM.get_visualnet_tiles(network)
 
-
+/obj/effect/psychic_tracer/Destroy()
+	EM = null //Clears out the ref that's about to become nulled to save GC time.
+	. = ..()
 
 /*
 	Extension: Added to the infected mob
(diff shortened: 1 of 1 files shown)


code/modules/mob/observer/freelook/marker/signal_abilities/psychic_tracer.dm
```diff
@@ -133,11 +135,11 @@
 	remove_extension(holder, base_type)
 
 /datum/extension/psychic_tracer/Destroy()
-	QDEL_NULL(object)
```
NanakoAC: QDEL_NULL isnt a hard delete, it just calls qdel, its right there in the name.  Its a macro and what you seem to have done here is broken it back up into the two lines it calls

NanakoAC (changes requested):
i dont think this change does what you think it does

NanakoAC (approved):

PR #1207, opened 2021-01-21
Better Non lethal - Taser

Sibg:
Gives the Taser a minor buff needed to get through rig armour, and a lore change to winchester arms.

### DIFF
--- a/code/modules/projectiles/projectile/beams.dm
+++ b/code/modules/projectiles/projectile/beams.dm
@@ -138,9 +138,8 @@
 	name = "stun beam"
 	icon_state = "stun"
 	fire_sound = 'sound/weapons/Taser.ogg'
-	check_armour = "energy"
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/projectile/beams.dm
```diff
@@ -138,9 +138,8 @@
 	name = "stun beam"
 	icon_state = "stun"
 	fire_sound = 'sound/weapons/Taser.ogg'
-	check_armour = "energy"
 	sharp = 0 //not a laser
-	agony = 40
```
NanakoAC, line 143: this is a bit larger than we'd usually buff in a single step, go with 45 for now

NanakoAC (changes requested):
Our balancing policy is incremental. Generally make changes 10-15% at a time

NanakoAC (approved):

PR #1232, opened 2021-02-03
Advanced Shield

JakeDaBoss:


code/game/objects/items/weapons/shields.dm
```diff
@@ -84,6 +84,22 @@
 	resistance = 6
 	blocksound = 'sound/items/trayhit2.ogg'
 
+/obj/item/weapon/shield/riot/advanced
```
NanakoAC: This might be a bit much. 

I'm assuming these are using the new sprites from imperator? I had intended to just replace the existing riot shield sprites with them, but a fancy version isn't a bad idea

Some concerns here:
1. Is the metal riot shield even obtainable ingame? I don't think so. And if not, it might be best to just delete that and use its stats here, i already designed and tested those
2. This ought to be added to random loot lists (along with the normal shield)

JakeDaBoss: I'll work on that. I have to doublecheck with whoever uses R&D or Security a lot, so I can cross-ref if they're even obtainable.

PR #1276, opened 2021-02-12
Fixed versioning of RIG Selection Overhaul

JakeDaBoss:
* Fixed an issue where nobody could select RIG modules anymore.


NanakoAC:
What's going on with this?

JakeDaBoss:
> What's going on with this?

I have quite a few things on the docket currently, meaning I'll get to this, but it's a lower priority than before.

PR #1279, opened 2021-02-14
Round end crew check tweak

### LINKED ISSUE #1114: Unbalanced counting of alive humans that are not on the ship side of the map.
Description of issue: Players on Aegis are counted towards alive crew. This causes stalemates, 6 rounds out of 10.

SSD players are likewise counted, and to my knowledge, alive but ghosted mobs too.

Difference between expected and actual behavior:
To prevent prolonged rounds of boredom of hunting down that last SSD/Ghosted mob, do not count them as alive in calcs for end of round.

Same for those on Aegis. At round end, any survivor on Aegis is considered Marooned and should not be counted.

Steps to reproduce:

Kill all alive humans on Ishimura.
See what happens if there are alive folks on the colony, or SSD/Ghosted mobs are still around.

JeserTheFox:
tweak: "Round end check for alive crewmates changed: only alive active humans on Ishimura count now."

Closes: https://github.com/DS-13-Dev-Team/DS13/issues/1114

### DIFF
--- a/code/_helpers/crew.dm
+++ b/code/_helpers/crew.dm
@@ -2,16 +2,9 @@
 /proc/get_historic_crew_total()
 	return GLOB.all_crew.len
 
-//Returns the number of alive humans aboard Ishimura
+//Returns the number of live crewmembers
 /proc/get_living_crew_total()
-	var/crew_count = 0
-	for(var/datum/mind/M in GLOB.living_crew)
-		if(!M || QDELETED(M.current) || !ishuman(M.current))	//regular sanity checks
-			continue
-		var/mob/living/L = M.current
-		if(!L.client || L.stat == DEAD || !isStationLevel(L.z))	//only count alive active humans onboard Ishimura, the rest are marooned
-			continue
-		crew_count++
+	return GLOB.living_crew.len
 
 	return crew_count
 
--- a/code/_helpers/crew.dm
+++ b/code/_helpers/crew.dm
@@ -2,10 +2,18 @@
 /proc/get_historic_crew_total()
 	return GLOB.all_crew.len
 
-//Returns the number of live crewmembers
+//Returns the number of alive humans aboard Ishimura
 /proc/get_living_crew_total()
-	return GLOB.living_crew.len
+	var/crew_count = 0
+	for(var/datum/mind/M in GLOB.living_crew)
+		if(!M || QDELETED(M.current) || !ishuman(M.current))	//regular sanity checks
+			continue
+		var/mob/living/L = M.current
+		if(!L.client || L.stat == DEAD || !isStationLevel(L.z))	//only count alive active humans onboard Ishimura, the rest are marooned
+			continue
+		crew_count++
 
+	return crew_count
 
 /*
 	Returns a list of all living crewmembers whose role fits the filter
--- a/code/_helpers/crew.dm
+++ b/code/_helpers/crew.dm
@@ -6,8 +6,6 @@
 /proc/get_living_crew_total()
 	return GLOB.living_crew.len
 
-	return crew_count
-
 /*
 	Returns a list of all living crewmembers whose role fits the filter
 	If role type is true, checks special role (AKA, antag status)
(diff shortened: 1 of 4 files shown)


code/_helpers/crew.dm
```diff
@@ -2,6 +2,15 @@
...
-//Returns the number of live crewmembers
+//Returns the number of alive humans aboard Ishimura
 /proc/get_living_crew_total()
-	return GLOB.living_crew.len
+	var/crew_count = 0
+	for(var/datum/mind/M in GLOB.living_crew)
+		if(!M || QDELETED(M.current) || !ishuman(M.current))	//regular sanity checks
+			continue
+		var/mob/living/L = M.current
+		if(!L.client || L.stat == DEAD || !isStationLevel(L.z))	//only count alive active humans onboard Ishimura, the rest are marooned
```
NanakoAC: hmm, on second thought, modifying this proc isnt a good idea, make a wrapper which calls it

Also, afaik, people are automatically removed from the list if killed/deleted, so those checks are unnecessary. likewise with the human check

Make a wrapper version that just checks station level and client
Also consider using ```client.is_afk(time)``` to check if people are sitting idle while still connected

NanakoAC (changes requested):

JakeDaBoss (changes requested):
Please consider making the following changes:
- Timer check.
- Checks if players are on Ishimura BUT ALSO on actual turfs. If space turf, consider them not on the ship?

code/game/gamemodes/gamemode.dm
```diff
@@ -650,3 +650,15 @@ proc/get_nt_opposed()
 			if (L.stat != DEAD) //They're alive!
 				GLOB.living_crew |= M
 
+//proc returns the amount of alive active humans onboard Ishimura (space turfs are excluded), the rest are considered marooned
+/datum/game_mode/proc/get_alive_crewmates_count()
+	var/crew_count = 0
+	for(var/datum/mind/M in GLOB.living_crew)
+		var/mob/living/L = M.current
+		if(!L.client || L.client.is_afk(2 MINUTES))	//activity check
```
JakeDaBoss, line 662: I'm going to leave this comment for posterity, but 2 minutes might be low. Only way we'll know is by testing.

I'm saying this because sometimes humans have to hide, be silent, and don't move as to not make any sound.

JakeDaBoss (approved):
Nice work!

code/game/gamemodes/gamemode.dm
JeserTheFox: Well, I figured that 1 minute will indeed be too low, however two minutes is enough, since it's quite a lot of time in terms of movement or doing anything in game.

JakeDaBoss: Fair enough. Thanks for your work!

code/game/gamemodes/marker/gamemodes.dm
```diff
@@ -143,7 +143,7 @@ Non-critical characters like any ghost-roles you may wish to add, or even antags
 	if(marker_active)	//Marker must be active
 		if (get_historic_crew_total() >= minimum_historic_crew)	//We need to have had a minimum total crewcount
 			var/minimum_living_crew = Ceiling(get_historic_crew_total() * minimum_alive_percentage)	//This many crew players at least, need to be left alive
-			if (get_living_crew_total() < minimum_living_crew)
+			if (get_alive_crewmates_count() < minimum_living_crew)
```
NanakoAC: Everything is good except the function name, its ambiguous and too similar to the former.

call it get_living_active_crew_aboard_ship
Its a bit of a mouthful but it'd be accurate

NanakoAC (changes requested):

code/game/gamemodes/marker/gamemodes.dm
JeserTheFox: Welp, I was trying to keep it somewhat compact. That was initial name for it, heh.

NanakoAC (approved):

PR #1291, opened 2021-02-18
Fixes maps in Run Linters

DeadLineSS13:
## About this PR
This PR will fix run_linters in maps.


NanakoAC (approved):
After discussion, approving this. But if we run into insurmountable problems in future with merging the tram work, this PR will need to be overwritten and redone.

The things changed here are easy to redo with converter tools so shouldnt be an issue

PR #1296, opened 2021-02-21
fix: Necromorph Limbs can be attached to humans.

### LINKED ISSUE #1139: Necromorph Limbs can be attached to players.
#### Description of issue:
people can de limb their own arm and attach a necro bladed one towards their own shoulder stump and vice versa can happen towards a necro.

Should not be able to attach these. It is however possible due to being a subtype of humans.

#### Steps to reproduce:
Delimb own arm, try to attach a necro limb.

Evankhell561:
fix: #1139
:cl:
fix: Necromorph Limbs can be attached to humans.
/:cl:

### DIFF
--- a/code/modules/surgery/limb_reattach.dm
+++ b/code/modules/surgery/limb_reattach.dm
@@ -38,7 +38,12 @@
 		if(!BP_IS_ROBOTIC(using))
 			to_chat(user, "<span class='danger'>You cannot attach a flesh part to a robotic body.</span>")
 			return SURGERY_FAILURE
-	return 1
+	if(istype(tool, /obj/item/organ/external))
+		var/obj/item/organ/external/E = tool
+		if(istype(E.species, /datum/species/necromorph) && !istype(target.species, /datum/species/necromorph))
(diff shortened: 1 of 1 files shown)


NanakoAC (changes requested):
This doesn't seem right, specific species typechecks here are generally a violation of good design

If preventing the use of necromorph limbs is the goal, the best way to do it would probably be horrible infections for the host. Outright preventing it like this feels arbitrary, plus its something players love to do for fun

Evankhell561:
> This doesn't seem right, specific species typechecks here are generally a violation of good design
> 
> If preventing the use of necromorph limbs is the goal, the best way to do it would probably be horrible infections for the host. Outright preventing it like this feels arbitrary, plus its something players love to do for fun

I understand I think the problem is that the sprites do not fit in the human sprite, a better solution (to begin with) would be to make sprites of the arms of the necromorphs that fit in humans.

NanakoAC (approved):
ah bother, you're right, preventing that kind of bugginess takes precedence, lets go with this for now

PR #1297, opened 2021-02-21
fix Line Guns

### LINKED ISSUE #1150: Line Guns have a single tile worth of "dead space".
Description of issue: Line guns have a tile of dead space (hah) between them and the actual projectile, where people are safe.

Difference between expected and actual behavior: You shouldn't be in melee range with someone and be safe when they shoot a line gun at you. But that is the case. There's a tile of safe space before the line gun's projectiles appear, and I've seen it used in-game in firing lines purposefully, even though it's a bug.

Steps to reproduce: Stand face to face with someone and shoot past them with a line gun. It won't hurt them.

Evankhell561:
fix: #1150
:cl:
fix: Line Guns have a single tile worth of "dead space". 
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/ds13/line_cutter.dm
+++ b/code/modules/projectiles/guns/ds13/line_cutter.dm
@@ -14,7 +14,6 @@
 	icon_state = "linecutter"
 	item_state = "linecutter"
 
-	ammo_type = /obj/item/ammo_casing/linerack
 	slot_flags = SLOT_BACK
 	mag_insert_sound = 'sound/weapons/guns/interaction/line_magin.ogg'
 	mag_remove_sound = 'sound/weapons/guns/interaction/line_magout.ogg'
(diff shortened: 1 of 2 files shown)


code/modules/projectiles/guns/ds13/line_cutter.dm
```diff
@@ -14,7 +14,6 @@
 	icon_state = "linecutter"
 	item_state = "linecutter"
 
-	ammo_type = /obj/item/ammo_casing/linerack
```
JakeDaBoss, line 17: This seems to be removed. Why is that?

Evankhell561: duplicate var:
image

Evankhell561: image

Evankhell561: even so, I had forgotten this change, I did it when I was trying to apply this fix in a different way.

JakeDaBoss (approved):

NanakoAC (approved):
sorry for delay, wanted to test this first, done so

PR #1311, opened 2021-03-07
opening the game panel no longer slows the game down

Evankhell561:
opening the game panel no longer automatically displays all available objects.
displaying all objects in the game consumes a lot of resources and slows down the game for a few seconds.
:cl:
tweak: opening the game panel no longer automatically displays all available objects.
/:cl:

### DIFF
--- a/html/create_object.html
+++ b/html/create_object.html
@@ -14,8 +14,8 @@
       color: #00f;
       font-family: Georgia, Arial, sans-serif;
     }
-    img { 
-      border: 0px; 
+    img {
+      border: 0px;
     }
     p.lic {
       font-size: 6pt;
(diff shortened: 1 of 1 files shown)


html/create_object.html
```diff
@@ -46,20 +46,19 @@
...
+
 	<script language="JavaScript">
 		var old_search = "";
 		var object_list = document.spawner.object_list;
 		var object_paths = null /* object types */;
 		var objects = object_paths == null ? new Array() : object_paths.split(";");
-		
+
 		document.spawner.filter.focus();
-		populateList(objects);
```
Evankhell561, line 57: this is the only change made.

NanakoAC: this requires explanation. What exactly does this do and why remove it?

Evankhell561: when opening the game panel, this automatically causes the list of all objects in the game to be displayed.

this happens for a second during which the game freezes and at the end, you don't search within that list but filter the path you want to find via the search engine.

Evankhell561: In general, this makes the processor cry every time the game panel is opened.

PR #1314, opened 2021-03-08
Security Changes

Sibg:
Flashes, Stun guns, Rig flashes, Stun guns have been removed due to opposite use of what they were intended for, stunning criminals not necromorths.


NanakoAC (approved):
flashes will be fixed in future, and we'll be adding the bolas function for shotgun soon too, so this is fine for now

PR #1317, opened 2021-03-12
SCL Shotgun

Evankhell561:
:cl:
add: SCL shotgun
add: magazine SCL-shotgun shells
/:cl:


NanakoAC:
Ah theres been a bit of a miscommunication, we eventually agreed on having the bolas as an alt fire, not an alternate ammo type. 
One kind of ammo, two firemodes.

We have lots of other examples of this, check pulse rifle and line cutter for demonstraton of alternate projectiles

code/game/objects/items/weapons/legcuffs.dm
```diff
@@ -40,18 +40,9 @@
 /obj/item/weapon/legcuffs/bola/throw_impact(atom/hit_atom)
 	if(..() || !ishuman(hit_atom))//if it gets caught or the target can't be cuffed,
 		return//abort
-	var/mob/living/carbon/human/H = hit_atom
-	/*
-	var/num_legs
+	cacht(hit_atom)
```
NanakoAC: cacht? I'm thinking you mean catch, although proc names should probably be more descriptive

code/modules/projectiles/ammunition/bullets.dm
```diff
@@ -144,6 +144,13 @@
 	projectile_type = /obj/item/projectile/energy/flash/flare
 	matter = list(MATERIAL_STEEL = 90, MATERIAL_GLASS = 90)
 
+/obj/item/ammo_casing/shotgun/bola
```
NanakoAC: this shouldn't be needed

code/modules/projectiles/guns/ballistic.dm
```diff
@@ -188,6 +188,15 @@
 	if(handle_casings != HOLD_CASINGS)
 		chambered = null
 
+/obj/item/weapon/gun/projectile/proc/load_magazine(obj/item/A, mob/user)
```
NanakoAC: what have you done with this proc?

code/modules/projectiles/guns/projectile/shotgun.dm
```diff
@@ -124,3 +124,39 @@
...
+	magazine_type = /obj/item/ammo_magazine/shotgun
+	var/aditional_magazine = /obj/item/ammo_magazine/shotgun
+	w_class = ITEM_SIZE_NORMAL
+	load_method = MAGAZINE
+	caliber = "shotgun"
+	force = 5
+	one_hand_penalty = 0
+	fire_delay = 40
+	firemodes = list(
+		list(mode_name = "sotgun", fire_delay = 2.5 SECONDS),
```
NanakoAC: "shotgun" and "bolas"

NanakoAC: 2.5 second firing delay seems really excessive, make it 1 second

and where a variable is shared across all firemodes you can just set it on the gun directly

code/game/objects/items/weapons/legcuffs.dm
```diff
@@ -40,18 +40,9 @@
...
 
-
-	for(var/limb_tag in H.species.bola_limbs)
-		var/obj/item/organ/external/E = H.has_organ_or_replacement(limb_tag)
-		if(E)
-			num_legs++
-
-	if(!H.legcuffed && num_legs >= 2)
-	*/
+/obj/item/weapon/legcuffs/bola/proc/cacht(var/mob/living/carbon/human/H)
```
NanakoAC: I dont see anything in here that determines escape time. A bola should be relatively easy to escape from, maybe 5 seconds of resisting.

Also does this work on necromorphs? It should work on any that have a bipedal form and both limbs, you may need to edit the hud data for necros to allow it, check the lurker for an example

NanakoAC (changes requested):
Several requested changes as outlined in comments, and here's a checklist of things too:
-Bolas should work on any bipedal creature with both of its legs, which is standing up
-They should fail when the victim is lying, missing a leg, or has more than two legs
-They should fail on creatures which aren't medium sized
-They should knock down charging necromorphs mid charge (spawn a charger to test this)
-Should take 5-10 seconds of resisting to escape

code/game/objects/items/weapons/legcuffs.dm
ghost: It works on necromorphs.

code/modules/projectiles/guns/ballistic.dm
Evankhell561: this made the previous version of the shotgun work.

code/game/objects/items/weapons/legcuffs.dm
Evankhell561: image

NanakoAC (approved):

PR #1344, opened 2021-04-01
Refinery Changes

Sibg:
The mining refinery has been moved to aegis for mining purposes and makes it easier to process and ship to the ishimura.


NanakoAC:
seems like a good change, but merging this will be on hold until ADS is merged, due to its largescale changes to the size of many maps. After that, the changes aboard ishimura will need redone

Issue #1218, opened 2021-01-27
Suggestion: Change access on Executive Freezer Fridges


atenistra:

Snypehunter007:
Seems like an oversight on my part when I was originally doing this.

Accepted, @atenistra

PR #1345, opened 2021-04-02
Change access on Executive Freezer Fridges #1218

Sibg:
Fixes secure fridges, you know.


code/game/objects/structures/crates_lockers/closets/secure/freezer.dm
```diff
@@ -5,7 +5,7 @@
 	icon_locked = "fridge1"
 	icon_opened = "fridgeopen"
 	icon_off = "fridgebroken"
-	req_access = list(access_service)
```
NanakoAC, line 8: Wrong way to do it.
This list requires the user to have ALL of these accesses to open it.

I believe you want to change it to req_one_access which allows access if you have anything in the list

NanakoAC (changes requested):

PR #1346, opened 2021-04-02
Make config as subsystem.

DeadLineSS13:
## About this PR
Add helpers qdell
Rework config (thx TG) as subsystem


code/__defines/admin.dm
```diff
@@ -39,6 +39,30 @@
 
 #define R_MAXPERMISSION 0x8000 // This holds the maximum value for a permission. It is used in iteration, so keep it updated.
 
+#define ADMIN_QUE(user) "(<a href='?_src_=holder;[HrefToken(TRUE)];moreinfo=[REF(user)]'>?</a>)"
```
NanakoAC: all these shortenings and abbreviations are no good, use full titles in the defines. self documenting code

code/__defines/admin.dm
```diff
@@ -39,6 +39,30 @@
...
+#define ADMIN_PP(user) "(<a href='?_src_=holder;[HrefToken(TRUE)];playerpanel=[REF(user)]'>PP</a>)"
+#define ADMIN_VV(atom) "(<a href='?_src_=vars;[HrefToken(TRUE)];vars=[REF(atom)]'>VV</a>)"
+#define ADMIN_SM(user) "(<a href='?_src_=holder;[HrefToken(TRUE)];subtlemessage=[REF(user)]'>SM</a>)"
+#define ADMIN_TP(user) "(<a href='?_src_=holder;[HrefToken(TRUE)];traitorpanel=[REF(user)]'>TP</a>)"
+#define ADMIN_KICK(user) "(<a href='?_src_=holder;[HrefToken(TRUE)];kick=[REF(user)]'>KICK</a>)"
+#define ADMIN_SC(user) "(<a href='?_src_=holder;[HrefToken(TRUE)];spawncookie=[REF(user)]'>SC</a>)"
+#define ADMIN_SFC(user) "(<a href='?_src_=holder;[HrefToken(TRUE)];spawnfortunecookie=[REF(user)]'>SFC</a>)"
+#define ADMIN_LOOKUP(user) "[key_name_admin(user)][ADMIN_QUE(user)]"
+#define ADMIN_LOOKUPFLW(user) "[key_name_admin(user)][ADMIN_QUE(user)] [ADMIN_FLW(user)]"
+#define ADMIN_FULLMONTY_NONAME(user) "[ADMIN_QUE(user)] [ADMIN_PP(user)] [ADMIN_VV(user)] [ADMIN_SM(user)] [ADMIN_JMP(user)] [ADMIN_FLW(user)]"
```
NanakoAC: get rid of all this monty stuff and name these defines properly according to what they do.

i dont know if monty is some TG inside joke, or the name of a developer, but it means nothing to us

code/__defines/configuration.dm
```diff
@@ -0,0 +1,7 @@
+//config files
+#define CONFIG_GET(X) global.config.Get(/datum/config_entry/##X)
```
NanakoAC, line 2: make these GET_CONFIG and SET_CONFIG, mass replace everywhere

code/_helpers/lists.dm
```diff
@@ -895,3 +895,31 @@ proc/dd_sortedTextList(list/incoming)
 
 
 #define IS_VALID_INDEX(list, index) (list.len && index > 0 && index <= list.len)
+
+//Like typesof() or subtypesof(), but returns a typecache instead of a list
+/proc/typecacheof(path, ignore_root_path, only_root_path = FALSE)
```
NanakoAC, line 900: what exactly is the point of this? it seems to just create a big nested list of path = TRUE
how does that help anything?
what is a typecache

code/_helpers/game.dm
```diff
@@ -359,5 +359,5 @@ datum/projectile_data
 /proc/SecondsToTicks(var/seconds)
 	return seconds * 10
 
-/proc/round_is_spooky(var/spookiness_threshold = config.cult_ghostwriter_req_cultists)
+/proc/round_is_spooky(var/spookiness_threshold = CONFIG_GET(number/cult_ghostwriter_req_cultists))
```
NanakoAC, line 362: does this even compile?

code/_helpers/logging.dm
```diff
@@ -38,11 +43,11 @@
 
 /proc/log_admin(text)
 	GLOB.admin_log.Add(text)
-	if (config.log_admin)
+	if (CONFIG_GET(flag/log_admin))
```
NanakoAC, line 46: flag/log_admin ? What is that, doesnt seem like valid syntax

code/_helpers/type2type.dm
```diff
@@ -232,3 +232,39 @@
 /proc/atomtype2nameassoclist(var/atom_type)
 	return atomtypes2nameassoclist(typesof(atom_type))
 
+//returns a string the last bit of a type, without the preceeding '/'
+/proc/type2top(the_type)
+	//handle the builtins manually
+	if(!ispath(the_type))
+		return
+	switch(the_type)
```
NanakoAC, line 240: this really doesnt seem like it should work

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+
+	var/resident_file	//the file which this was loaded from, if any
+	var/modified = FALSE	//set to TRUE if the default has been overridden by a config entry
+
+	var/deprecated_by	//the /datum/config_entry type that supercedes this one
+
+	var/protection = NONE
+	var/abstract_type = /datum/config_entry	//do not instantiate if type matches this
+
+	var/vv_VAS = TRUE		//Force validate and set on VV. VAS proccall guard will run regardless.
```
NanakoAC: whats this VAS thing?

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+	var/key_mode
+	var/value_mode
+	var/splitter = " "
+
+/datum/config_entry/keyed_list/New()
+	. = ..()
+	if(isnull(key_mode) || isnull(value_mode))
+		CRASH("Keyed list of type [type] created with null key or value mode!")
+
+/datum/config_entry/keyed_list/ValidateAndSet(str_val)
```
NanakoAC, line 125: this whole file is badly in need of comments, especially this proc

NanakoAC (changes requested):
see comments. 
review not finished yet, i only got as far as  code/controllers/configuration/configuration.dm

will continue later, need sleep

code/_helpers/_qdel.dm
```diff
@@ -0,0 +1,20 @@
+#define QDELING(X) (X.gc_destroyed)
+#define QDELETED(X) (!X || QDELING(X))
+#define QDESTROYING(X) (!X || X.gc_destroyed == GC_CURRENTLY_BEING_QDELETED)
+
+#define QDEL_IN(item, time) addtimer(CALLBACK(GLOBAL_PROC, .proc/qdel, item), time, TIMER_STOPPABLE)
+#define QDEL_IN_CLIENT_TIME(item, time) addtimer(CALLBACK(GLOBAL_PROC, .proc/qdel, item), time, TIMER_STOPPABLE | TIMER_CLIENT_TIME)
+#define QDEL_NULL(item) qdel(item); item = null
```
Evankhell561, line 7: #define QDEL_NULL(x) if(x) { qdel(x) ; x = null }

Evankhell561 (changes requested):

code/_helpers/_qdel.dm
DeadLineSS13: > #define QDEL_NULL(x) if(x) { qdel(x) ; x = null }

This define deleted from defines/qdel.dm

code/_helpers/game.dm
DeadLineSS13: > does this even compile?

Yep

code/_helpers/logging.dm
DeadLineSS13: > flag/log_admin ? What is that, doesnt seem like valid syntax

This bool var for log admin.

code/_helpers/type2type.dm
DeadLineSS13: > this really doesnt seem like it should work

But it also works fine.

code/_helpers/_qdel.dm
Evankhell561: I prefer the old definition. why change it?

code/controllers/configuration/config_entry.dm
DeadLineSS13: > whats this VAS thing?

I forget delete this var because this var for optimized VV TG.

code/_helpers/_qdel.dm
DeadLineSS13: > I prefer the old definition. why change it?

Because I extended the Qdel functions and removed the extra characters? For you, the value of the returned action will not change.

code/__defines/configuration.dm
DeadLineSS13: > make these GET_CONFIG and SET_CONFIG, mass replace everywhere

I think that in syntaxis better CONFIG_GET, because CONFIG is system GET is proc. You don't write destroy.mob in the code.

code/controllers/configuration/config_entry.dm
NanakoAC: comments! At least explain what these procs do

code/controllers/configuration/entries/game_option.dm
```diff
@@ -0,0 +1,117 @@
...
+
+/datum/config_entry/flag/organs_decay
+
+//Paincrit knocks someone down once they hit 60 shock_stage, so by default make it so that close to 100 additional damage needs to be dealt,
+//so that it's similar to PAIN. Lowered it a bit since hitting paincrit takes much longer to wear off than a halloss stun.
+/datum/config_entry/number/organ_damage_spillover_multiplier
+    config_entry_value = 0.5
+    integer = FALSE
+
+/datum/config_entry/flag/bones_can_break
```
NanakoAC, line 24: I believe we had bone breaking disabled

code/controllers/configuration/entries/game_option.dm
```diff
@@ -0,0 +1,117 @@
...
+
+/datum/config_entry/flag/generate_map
+
+/datum/config_entry/flag/no_click_cooldown
+
+//Used for modifying movement speed for mobs.
+//Unversal modifiers
+
+/datum/config_entry/number/run_speed
+    config_entry_value = 2
```
NanakoAC, line 53: Check carefully that these values match what we had before

code/controllers/configuration/entries/general.dm
```diff
@@ -0,0 +1,244 @@
...
+/datum/config_entry/flag/antag_hud_restricted   //Ghosts that turn on Antagovision cannot rejoin the round.
+
+/datum/config_entry/keyed_list/probabilities    //relative probability of each mode
+    key_mode = KEY_MODE_TEXT
+    value_mode = VALUE_MODE_TEXT
+
+/datum/config_entry/keyed_list/probabilities/ValidateListEntry(key_name)
+	return key_name in config.modes
+
+/datum/config_entry/flag/humans_need_surnames
```
NanakoAC, line 157: this should be TRUE

code/controllers/configuration/entries/general.dm
```diff
@@ -0,0 +1,244 @@
...
+
+/datum/config_entry/flag/allow_random_events    //enables random events mid-round when set to 1
```
NanakoAC, line 159: TRUE

DeadLineSS13: In config ready.

code/controllers/configuration/entries/general.dm
DeadLineSS13: Config too

code/controllers/configuration/entries/game_option.dm
DeadLineSS13: Config => game_options

code/controllers/verbs.dm
```diff
@@ -42,7 +42,7 @@
 			debug_variables(evacuation_controller)
 			feedback_add_details("admin_verb","DEmergency")
 		if("Configuration")
-			debug_variables(config)
+			debug_variables(global.config)
```
NanakoAC, line 45: why is this global.config instead of GLOB.config?

DeadLineSS13: > why is this global.config instead of GLOB.config?

Because GLOBAL_REAL(config, /configuration) <=> var/global/configuration/config

code/modules/admin/holder2.dm
```diff
@@ -52,7 +55,24 @@ var/list/admin_datums = list()
 		owner.deadmin_holder = null
 		owner.add_admin_verbs()
 
-
+/proc/GenerateToken()
```
NanakoAC: what are these tokens for?

code/modules/admin/view_variables/view_variables_global.dm
```diff
@@ -29,14 +29,6 @@
 				"forumsqllogin",
 				"forumsqlpass",
 				"forumsqlport",
-				"sqladdress",
```
NanakoAC, line 32: why are these all deleted? we plan to use an sql database in future

code/modules/admin/holder2.dm
DeadLineSS13: It is for HrefToken, but I think we don't need HrefToken at now moment.

code/modules/admin/view_variables/view_variables_global.dm
DeadLineSS13: > why are these all deleted? we plan to use an sql database in future

Because Sqladress now in Config. No shitty changes

config/example/config.txt
```diff
@@ -282,33 +283,34 @@ USEALIENWHITELIST
...
-## Sets the number of loadout slots per character
-LOADOUT_SLOTS 3
-
 ## Expected round length in minutes
 EXPECTED_ROUND_LENGTH 180
 
 ## The lower delay between events in minutes.
 ## Affect mundane, moderate, and major events respectively
-EVENT_DELAY_LOWER 10;30;50
+EVENT_DELAY_LOWER 1 10
```
NanakoAC: I need these event variables moved out of config entirely, make them hardcoded defines instead

NanakoAC:
why is debug config deleted?

code/controllers/configuration/entries/game_option.dm
NanakoAC: Right, but you can set a default option here can't you? Just in case something goes wrong with config

same for the other two comments

NanakoAC:
is this ready to go? comments addressed?

DeadLineSS13:
> is this ready to go? comments addressed?

No ready. Because there have problem keyd list

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+	if(!VASProcCallGuard(str_val))
```
DeadLineSS13, line 126: Protect calls

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+	if(key_pos || value_mode == VALUE_MODE_FLAG)
+		key_name = lowertext(copytext(str_val, 1, key_pos))
+		if(key_pos)
+			key_value = copytext(str_val, key_pos + length(str_val[key_pos]))
+		var/new_key
+		var/new_value
+		var/continue_check_value
+		var/continue_check_key
+		switch(key_mode)
+			if(KEY_MODE_TEXT)
```
DeadLineSS13, line 143: If key name text

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+				new_key = key_name
+				continue_check_key = new_key
+			if(KEY_MODE_TYPE)
```
DeadLineSS13, line 146: If key name path

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+				new_key = key_name
+				if(!ispath(new_key))
+					new_key = text2path(new_key)
+				continue_check_key = ispath(new_key)
+		switch(value_mode)
+			if(VALUE_MODE_FLAG)
```
DeadLineSS13, line 152: List with FLAGs

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+				new_value = TRUE
+				continue_check_value = TRUE
+			if(VALUE_MODE_NUM)
```
DeadLineSS13, line 155: List with numbers

code/controllers/configuration/config_entry.dm
```diff
@@ -0,0 +1,166 @@
...
+				new_value = text2num(key_value)
+				continue_check_value = !isnull(new_value)
+			if(VALUE_MODE_TEXT)
```
DeadLineSS13, line 158: List with text

DeadLineSS13 (commented):
Work ValidateAndSet

DeadLineSS13:
> why is debug config deleted?

Fixed. (I don't know how ignores these files)

NanakoAC:
> 
> 
> > why is debug config deleted?
> 
> Fixed. (I don't know how ignores these files)

gitignore causes this, you have to make your git client show ignored files and manually stage them

code/game/gamemodes/gameticker.dm
```diff
@@ -99,13 +99,8 @@ var/global/datum/controller/gameticker/ticker
 	else
 		src.hide_mode = 0
 
-	if((master_mode=="random") || (master_mode=="secret"))
```
NanakoAC, line 103: what is all this you're doing with gamemodes?

DeadLineSS13: > what is all this you're doing with gamemodes?

At the moment, you are running the voting mode at startup. And since probality needs to be reworked into a normal form (in the new version of the probality configs, they do not work as needed). So I decided to remove this so that we wouldn't experience such a problem.

NanakoAC (approved):

PR #1358, opened 2021-04-10
fix vector code.

Evankhell561:
the vector code was not written in a proper byond format and was causing some problems.
from:
image

to:
image


lib/vector2/CommonVectors.dm
```diff
@@ -2,7 +2,7 @@
 	Make sure you don't modify these vectors.
 	You should treat all vectors as immutable in general.
 */
-var Vector2/Vector2 = new
+var/global/Vector2/Vector2 = new
```
DeadLineSS13: var/global bad idea. Use GLOBAL_REAL

DeadLineSS13 (changes requested):
This need, because you create new error in run linters with unmanaged global

lib/vector2/CommonVectors.dm
NanakoAC: this is used quite a lot as its a holder for functions, having to put GLOB infront of it would be a pain

DeadLineSS13: > this is used quite a lot as its a holder for functions, having to put GLOB infront of it would be a pain

GLOBAL_REAL don't need use GLOB. Check define GLOBAL_REAL

Evankhell561: if you could make a direct suggestion as to how I should write this line that would be great.

DeadLineSS13: > if you could make a direct suggestion as to how I should write this line that would be great.

GLOBAL_REAL(Vector2, /Vector2) = new

Evankhell561:
```suggestion
GLOBAL_REAL(Vector2, /Vector2) = new
```

NanakoAC (approved):

PR #1367, opened 2021-04-12
AdsPowerFix&BugFixes

### LINKED ISSUE #1161: Colony atmos meters measure room, not pipe
#### Description of issue:
Meters in colony's atmosphere room are turf-typed. They analyse the atmosphere at their turfs. Should be changed to not turf type.

See Desc.

### LINKED ISSUE #1218: Suggestion: Change access on Executive Freezer Fridges

### LINKED ISSUE #1292: Suggestion: Add Condimachine to kitchen
#### Suggestion:
Add a Condimachine to the kitchen.

This will allow the kitchen to make things that are in the wiki when they require separation of chemicals, like nutrient from rice.

Snypehunter007:
- Fixed the ADS room and airlock not getting power.
  - Fixed pipe meters on the Aegis Colony being the wrong type.
  - Removed the Service department access requirement for freezer lockers.
  - Added a Condimaster to the Messhall Kitchen.
  - Slightly re-arranged the ADS room and added a few extra supplies.
  
  Fixes #1292 
  Fixes #1161 
  Fixes #1218


NanakoAC (changes requested):
we need the TG map format in here, i've pinned a guide to doing it in the mapping channel, squad marine and draconic can also help if you get stuck

PR #1373, opened 2021-04-13
Arrival Hall Tileset

Ketrai:
First iteration of the arrival hall tiles. You can use them anywhere else, but it's based off arrival hall concept art.

NOTES FOR MAPPERS*

Slat decals note.
* only use with rivet flooring underneath
* c pieces always need to be combined with corner pieces
* ce and cs stand for start and end curve respectfully. ce comes before the corner piece, cs comes after the cornerpiece.
* twist pieces always need to be combined with corner pieces
* start building your slats from a slats_start piece, and end it at a slats_end piece
* pieces with a _r are reversed.

- Slashed pieces come with two borders left and right by default.

- Plastic slabs s (small) and m (medium) can be combined on the same tile for effect.

- rivet flooring is meant to be placed under slat pieces, it's made with a simple texture in mind as to avoid making slat tiles too busy.


NanakoAC:
do you still want this in? it seemed like you were still tweaking and testing things out when we last discussed it

NanakoAC:
@Ketrai are you done with this?

PR #1391, opened 2021-04-24
add Brace

Evankhell561:
adds an object that blocks the doors in one direction.
image
image

:cl:
add: airlock directional brace
/:cl:


NanakoAC:
this just looks like a bolted door, not distinct enough.
I had a set of sprites made for this

code/game/machinery/doors/airlock.dm
```diff
@@ -90,6 +90,9 @@ var/list/airlock_overlays = list()
 	var/welded_file = 'icons/obj/doors/station/welded.dmi'
 	var/emag_file = 'icons/obj/doors/station/emag.dmi'
 
+	/// Unrestricted sides. A bitflag for which direction (if any) can open the door with no access.
+	var/obj/item/airlock_reinforcement/reinforcement
```
NanakoAC: why would an obj/item var be used to hold a bitflag?

code/game/machinery/doors/airlock.dm
```diff
@@ -1512,4 +1533,35 @@ About the new airlock wires panel:
...
+		update_icon()
+		to_chat(M, SPAN_NOTICE("You apply [R] to [src]."))
+
+/obj/machinery/door/airlock/proc/remove_reinforce(obj/item/I, mob/M)
+	reinforcement.forceMove(get_turf(src))
+	icon = initial(icon)
+	update_icon()
+	to_chat(M, SPAN_NOTICE("You remove [reinforcement] from [src]."))
+
+/obj/item/airlock_reinforcement
```
NanakoAC: airlock braces already exist in baycode, this is redundant, may as well just repurpose them

NanakoAC (changes requested):
Reinforcing a door should add an extension to it to increase its resistance, as well as other changes listed here

code/game/machinery/doors/airlock.dm
Evankhell561: > airlock braces already exist in baycode, this is redundant, may as well just repurpose them

You are right, I realized this after I finished the code.

NanakoAC (changes requested):
A brace on an airlock should increase its resistance and force difficulty, to a degree that most weak necromorphs should be unable to open it. Balance it around the Brute

NanakoAC:
we also had our own sprite made for reinforcing airlocks

[metaldoor.zip](https://github.com/DS-13-Dev-Team/DS13/files/6489590/metaldoor.zip)

Evankhell561:
> A brace on an airlock should increase its resistance and force difficulty, to a degree that most weak necromorphs should be unable to open it. Balance it around the Brute

image
hunters cannot open using the grab intent.
as for the difficulty of destroying the door, when a door with a brace is hit, it redirects the damage to the brace. The brace itself usually has twice as much life as a normal door.

Evankhell561:
> we also had our own sprite made for reinforcing airlocks
> 
> [metaldoor.zip](https://github.com/DS-13-Dev-Team/DS13/files/6489590/metaldoor.zip)

these sprites are already in-game in this path:
'icons/obj/doors/station/metaldoor.dmi'. I already implemented it to my code.

NanakoAC:
ah my comment about the icon has been misunderstood, my bad, it wasn't the doors i wanted to use, but the metal bars, as the icon for the brace object

NanakoAC:
we also need a way to create it. Add a crafting recipe to make a brace, just a few metal sheets and some application of a welder, have a look at other crafting recipes for example

Evankhell561:
> ah my comment about the icon has been misunderstood, my bad, it wasn't the doors i wanted to use, but the metal bars, as the icon for the brace object

ok, ready but it does not fit much to the fact that the door can be opened in one direction.

image

Evankhell561:
> ah my comment about the icon has been misunderstood, my bad, it wasn't the doors i wanted to use, but the metal bars, as the icon for the brace object

ready

NanakoAC (approved):

PR #1394, opened 2021-04-27
more code fixes

Evankhell561:
from:
image

to:
image


code/_helpers/maths.dm
```diff
@@ -2,7 +2,7 @@
 #define RAND_F(LOW, HIGH) (rand()*(HIGH-LOW) + LOW)
 #define ceil(x) (-round(-(x)))
 #define CEILING(x, y) ( -round(-(x) / (y)) * (y) )
-#define NEAR_INFINITY	9999999999	//Just a really big number
+#define NEAR_INFINITY	1000000000	//Just a really big number
```
DeadLineSS13, line 5: I think it is better to do the value in the style of this #define INFINITY	1e31

DeadLineSS13 (changes requested):
I think that this PR needs to be tested separately in a test run, since many important game mechanics are involved here.

Evankhell561:
> I think that this PR needs to be tested separately in a test run, since many important game mechanics are involved here.

a testmarge would be nice.

code/_helpers/maths.dm
Evankhell561: uhm, this is the largest integer that byond can support.

code/game/objects/structures/crates_lockers/closets/secure/secure_closets.dm
```diff
@@ -15,5 +15,5 @@
 	wall_mounted = 0 //never solid (You can always pass over it)
 	health = 200
 
-/obj/structure/closet/secure_closet/slice_into_parts(obj/item/weapon/weldingtool/WT, mob/user)
+/obj/structure/closet/secure_closet/slice_into_parts(obj/item/weapon/tool/weldingtool/WT, mob/user)//
```
Evankhell561:
```suggestion
/obj/structure/closet/secure_closet/slice_into_parts(obj/item/weapon/tool/weldingtool/WT, mob/user)
```

code/_macros.dm
```diff
@@ -79,8 +79,6 @@
 
 #define attack_animation(A) if(istype(A)) A.do_attack_animation(src)
 
-#define isairlock(A) istype(A, /obj/machinery/door/airlock)
```
NanakoAC: don't remove this, it'll be used in future, surprised its not already. There are plenty of places it could be

code/_onclick/click_mob.dm
```diff
@@ -324,10 +324,8 @@
 
 	while (click_handlers.Num())
 		var/datum/click_handler/CH = click_handlers.Pop()
-		if (CH)
-			if (!CH.OnAltClick(A,params))
-				return
-	.=..()
+		if(CH && !CH.OnAltClick(A,params))
```
NanakoAC: This change could break things, why is it done?

code/game/gamemodes/marker/marker_shard.dm
```diff
@@ -107,15 +107,15 @@
 /obj/item/marker_shard/proc/activate()
 	active = TRUE
 	last_known_location = loc
-	GLOB.moved_event.register(src, src, /obj/item/marker_shard/moved)
```
NanakoAC, line 110: why this? The left is how i've always done event registration, it needs some explanation

code/game/machinery/autolathe.dm
```diff
@@ -640,8 +640,8 @@
 
 
 	//And if there's any remainder, we eject that as a shard
-	if (remainder)
```
NanakoAC, line 643: this was valid before, shards contain a variable quantity of the material

besides this is going to be a non integer value in most cases, you can't iterate through it

code/game/machinery/computer/atmos_alert.dm
```diff
@@ -14,7 +14,7 @@ var/global/list/minor_air_alarms = list()
 
 /obj/machinery/computer/atmos_alert/Initialize()
 	. = ..()
-	atmosphere_alarm.register_alarm(src, /obj/machinery/computer/station_alert/update_icon)
```
NanakoAC, line 17: as with previous event change, this needs explanation. Why is it being changed to a different format than the previous one? why not...

```atmosphere_alarm.register_alarm(src, .proc/update_icon)```

File: code/game/mecha/mecha.dm [hunk 8]
```diff
@@ -1561,7 +1561,7 @@
 		var/mob/occupant = P.occupant
 
 		user.visible_message("<span class='notice'>\The [user] begins opening the hatch on \the [P]...</span>", "<span class='notice'>You begin opening the hatch on \the [P]...</span>")
-		if (!do_after(user, 40, needhand=0))
```
NanakoAC, line 1564: we've never used global.thing before, i don't see any reason to start doing it now

code/game/objects/items/stacks/stack.dm
```diff
@@ -84,7 +84,7 @@
 				continue
 			t1+="<br>"
 			var/max_multiplier = round(src.get_amount() / R.req_amount)
-			var/title as text
```
NanakoAC, line 87: was this invalid before?

code/game/objects/items/stacks/stack.dm
```diff
@@ -204,7 +204,6 @@
 			var/datum/matter_synth/S = synths[i]
 			S.use_charge(charge_costs[i] * used) // Doesn't need to be deleted
 		return 1
-	return 0
```
NanakoAC, line 207: whats wrong with a trailing return 0? isnt the default return value null?

code/modules/codex/entries/ascent.dm
```diff
@@ -4,7 +4,7 @@
 	associated_paths = list(
 		/mob/living/silicon/robot/flying/ascent,
 		/obj/item/device/multitool/mantid,
-		/obj/item/weapon/weldingtool/electric/mantid,
+		/obj/item/weapon/tool/weldingtool/electric/mantid,
```
NanakoAC, line 7: this doesnt exist, just delete the line

code/modules/economy/EFTPOS.dm
```diff
@@ -268,7 +268,5 @@
 				playsound(src, 'sound/machines/chime.ogg', 50, 1)
 				src.visible_message("\icon[src] \The [src] chimes.")
 				transaction_paid = 1
-	else
```
NanakoAC, line 271: calling parent in a trailing else seems valid, why remove this?

code/modules/hydroponics/trays/tray.dm
```diff
@@ -282,7 +282,7 @@
 	return TRUE
 
 //Clears out a dead plant.
-/obj/machinery/portable_atmospherics/hydroponics/proc/remove_dead(var/mob/user, var/silent)
```
NanakoAC: using var/ in proc parameters is normal and something we do everywhere. No purpose in creating isolated exceptions to established policy

code/modules/materials/materials.dm
```diff
@@ -187,7 +187,7 @@ var/list/name_to_material
 	if(islist(composite_material))
 		for(var/material_string in composite_material)
 			temp_matter[material_string] = composite_material[material_string]
-	else if(SHEET_MATERIAL_AMOUNT)
```
NanakoAC, line 190: this always returned true before, but it seems like the author may have intended something else

Have a look at surrounding code and see if this was a bug

code/modules/mob/living/abilities/gallop.dm
```diff
@@ -50,9 +50,8 @@
 		GLOB.moved_event.register(user, src, /datum/extension/gallop/proc/user_moved)
 
 /datum/extension/gallop/proc/destroy()
```
NanakoAC: i think i've messed up a bit here, this should probably be Destroy, and it should call parent

code/modules/mob/living/bot/mulebot.dm
```diff
@@ -208,7 +208,7 @@
 		M.Weaken(5)
 	..()
 
-/mob/living/bot/mulebot/proc/runOver(var/mob/living/carbon/human/H)
```
NanakoAC: isolated exception

code/modules/mob/living/carbon/human/species/necromorph/tripod.dm
```diff
@@ -858,6 +858,4 @@ https://bigmemes.funnyjunk.com/pictures/Long+boi_073bf6_7722185.jpg
 #undef LEAP_REDUCED_COOLDOWN
 #undef TONGUE_EXTEND_TIME
 
-#undef LEFT_ARM_OFFSETS
```
NanakoAC, line 861: why undef this?

code/modules/mob/living/carbon/human/species/necromorph/ubermorph.dm
```diff
@@ -277,7 +277,7 @@ Best used near the end, when all seems quiet, to help the necromorphs hunt down
 			//Find target organ should never fail, we won't bother checking
 			targetstring = "[H]'s [found_organ]"
 
-			playsound(charge., 'sound/weapons/slice.ogg', VOLUME_MAX, 1)
```
NanakoAC, line 280: this is a bug, the sound should be played at a variable of charge. User or target maybe, check into it

code/modules/projectiles/guns/special/ripper/sawblade.dm
```diff
@@ -279,16 +279,11 @@
...
 			qdel(src)
 
-#undef DROP_DAMAGE
 #undef IMPACT_DAMAGE
 
 #undef EX3_TOTAL
 #undef EX2_TOTAL
 #undef EX1_TOTAL
 
-#undef SOUND_GRIND
```
NanakoAC, line 289: why undef all of these?

code/modules/shuttles/shuttle_emergency.dm
```diff
@@ -8,10 +8,10 @@
 	emergency_controller = evacuation_controller
 	if(!istype(emergency_controller))
 		CRASH("Escape shuttle created without the appropriate controller type.")
-		return
```
NanakoAC, line 11: i dont see anything invalid about these returns

code/modules/tables/tables.dm
```diff
@@ -247,13 +247,13 @@
 	manipulating = 0
 	return null
 
-/obj/structure/table/proc/remove_reinforced(obj/item/weapon/screwdriver/S, mob/user)
```
NanakoAC, line 250: change the typepaths used in these parameters to just obj/item/weapon/tool
the subtypes are unnecessary as tool interactions are handled with qualities

NanakoAC (changes requested):
A lot of confusing points and changes that seem arbitrary

Several things need explanation

A few of the problems here are caused by underlying bugs that need fixed, and the proposed changed would just hide them. i've commented on those

One thing that came up repeatedly (removing leading var/) is out of place and should not be done piecemeal like this. If there's merit to doing it, that should be done seperately

code/_macros.dm
Evankhell561: this is duplicated.

code/game/mecha/mecha.dm
Evankhell561: the do_after proc is overwritten for this machine.

code/game/objects/items/stacks/stack.dm
Evankhell561: the "as text" has no use in local variables.

code/game/objects/items/stacks/stack.dm
Evankhell561: The problem here is that this coding line could never be reached. 
above this was:

```
if(something)

  code
  
  return
  
else

   code
   
   return
  
return
```

code/modules/economy/EFTPOS.dm
Evankhell561: this code was never reached.

code/modules/mob/living/carbon/human/species/necromorph/tripod.dm
Evankhell561: LEFT_ARM_OFFSETS
was never defined.

PR #1405, opened 2021-05-04
Necromorph Skill Change

ZanderDowdall:
This changes the default skill level from 4 (Expert) to 1 (None). This means necromorphs will no longer have expert level skill in all skills. This should fix the balance issue with necromorphs being able to spam disarm survivors, guaranteeing a kill.


ghost:
We have a PR up already for handling the disarm spam. Lowering skills across the board isn't the proper solution for something like this, and will result in Necromorphs missing attacks wildly. Closing this.

PR #1411, opened 2021-05-06
Shared cooldowns for snapshot and long shot

boskoramen:
This is per request. Also tweaked how shoot extensions handle base types in `can_shoot` so the shared cooldowns could work.

:cl: TheDonkified

add: Shared 4 second cooldowns for the snapshot and long shot abilities for the Spitter and Puker necromorphs.

/:cl:

### DIFF
--- a/code/modules/mob/living/abilities/shoot.dm
+++ b/code/modules/mob/living/abilities/shoot.dm
@@ -192,18 +192,22 @@
 /***********************
 	Safety Checks
 ************************/
-//Access Proc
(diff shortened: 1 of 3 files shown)


NanakoAC:
the original implementation is by design, so that playing these classes isn't simply spamming the same thing over and over

if these abilities are too powerful, they can be reduced, and if their individual cooldowns are too small they can be extended, but making them share a cooldown is a major violation of the design and actively removing gameplay

NanakoAC:
Supplanted by #1414

PR #1412, opened 2021-05-06
Crew Objective System: ADS

NanakoAC:
This PR implements our first proper crew objective. Some notes:
-The airlock to the ADS structure is still nonfunctional, snype will fix this
-To avoid staleness, the weighting for the ADS is relatively low, it will not automatically occur every round, or even most rounds. This will become less of an issue as we add more objectives in future. But if people want to see it urgently, an admin can trigger it anytime through the event manager panel
-This is a huge and complex PR, a closed test with staff is recommended to iron out issues


TerryTehCoder:
This is actually really awesome. Surprised there's not more feedback. Although, if the other crew objectives are anything like the ADS system, I can't imagine it's going to be easy to code them in.

PR #1416, opened 2021-05-09
New version observation

DeadLineSS13:
## About this PR
Now Observation is not global. This will optimize the Raise Event and the way observations are stored.


code/__defines/observation.dm
```diff
@@ -0,0 +1,7 @@
+/// Used to trigger observations and call procs registered for that observation
+/// The datum hosting the observation is automaticaly added as the first argument
+/// Returns a bitfield gathered from all registered procs
+/// Arguments given here are packaged in a list and given to _RaiseEvenet
+#define RAISE_EVENT(listener, obstype, arguments...) ( !listener.observations || !listener.observations[obstype.name] ? NONE : listener._RaiseEvent(obstype, list(listener, ##arguments)) )
```
NanakoAC: Listener is being used wrongly here, this is the source not the listener

NanakoAC: Don't take shortcuts with variable names, use observation_datum since what's being passed here is a datum ref, not a typepath

NanakoAC: Using names as list keys is not a good idea, nothing forces them to be unique.

Instead, use the observation datum itself as the list key, you can do that

code/datums/datum.dm
```diff
@@ -9,6 +9,21 @@
...
 	var/datum_flags = NONE
 
+	/**
+	  * Any datum registered to receive observations from this datum is in this list
+	  *
+	  * Lazy associated list in the structure of `observation:registree/list of registrees`
+	  */
+	var/list/observations
+	/// Lazy associated list in the structure of `observations:proctype` that are run when the datum receives that observation
+	var/list/list/datum/callback/observation_procs
```
NanakoAC: I really don't think this second list is necessary

code/datums/datum.dm
```diff
@@ -9,6 +9,21 @@
...
+	/**
+	  * Is this datum capable of sending observations?
+	  *
+	  * Set to true when a observation has been registered
+	  */
+	var/observation_enabled = FALSE
```
NanakoAC: this is redundant, remove it. The observations list being null or not is enough to determine if it cares about events

code/datums/datum.dm
```diff
@@ -31,6 +46,10 @@
 		if (timer.spent)
 			continue
 		qdel(timer)
+
+	for(var/target in observation_procs)
+		UnregisterObservation(target, observation_procs[target])
```
NanakoAC: Registering and unregistering should still be handled via the observation singletons

code/datums/observation/observation.dm
```diff
@@ -236,3 +236,113 @@
...
+ * the source datum this proc is called upon, will recieve a callback to the given proctype
+ * Return values from procs registered must be a bitfield
+ *
+ * Arguments:
+ * * datum/listener The listener to listen for observations from
+ * * obs_type_or_types Either a string observation name, or a list of observation names (strings)
+ * * proctype The proc to call back when the observation is emitted
+ * * override If a previous registration exists you must explicitly set this
+ */
+/datum/proc/RegisterObservation(datum/listener, obs_type_or_types, proctype, override = FALSE)
```
NanakoAC: as above, leave this to the singletons, they can just put data on the source datum as needed

code/datums/observation/observation.dm
```diff
@@ -236,3 +236,113 @@
...
+	if(QDELETED(src) || QDELETED(listener))
+		return
+
+	var/list/procs = observation_procs
```
NanakoAC: Don't do this. Making another variable which contains a list is meaningless

Lists are passed by reference, so any modification to this var will alter the original list too.

If that's not desireable, use list.Copy() to make a copy of it.
If that IS what you want to do then you don't need this extra var, just work on the original list

NanakoAC (changes requested):
several things

code/datums/datum.dm
NanakoAC: remove this

code/datums/datum.dm
NanakoAC: remove this

code/datums/datum.dm
NanakoAC: remove

code/controllers/observation.dm
```diff
@@ -0,0 +1,6 @@
+PROCESSING_SUBSYSTEM_DEF(observation)
```
NanakoAC: i dont see a need for this subsystem at all, get rid of it

code/datums/observation/observation.dm
```diff
@@ -121,15 +123,31 @@
 		callbacks = list()
 		listeners[listener] = callbacks
 
-	// If the proc_call is already registered skip
+	var/list/lookup = event_source.observations
```
NanakoAC: this serves no purpose, operate on the list directly

code/datums/observation/observation.dm
```diff
@@ -121,15 +123,31 @@
...
+		event_source.observations[src] = lookup[src] = list()
+	if(!lookup[src][listener])
+		event_source.observations[src][listener] = lookup[src][listener] = list()
+
+	var/list/procs = event_source.observations[src][listener]
+
+	if(!override && procs[event_source])
+		crash_with("[src] overridden. Use override = TRUE to suppress this warning")
+
+	procs[event_source] = proc_call
```
NanakoAC: just add the proc call don't key it to anything

code/__defines/observation.dm
```diff
@@ -0,0 +1,5 @@
+/// Used to trigger observations and call procs registered for that observation
+/// The datum hosting the observation is automaticaly added as the first argument
+/// Returns a bitfield gathered from all registered procs
+/// Arguments given here are packaged in a list and given to _RaiseEvenet
+#define RAISE_EVENT(event_source, observation_datum, arguments...) ( !event_source.observations || !event_source.observations[observation_datum] ? NONE : event_source.RaiseEvent(observation_datum, list(event_source, ##arguments)) )
```
NanakoAC: theres no need for datum.raise_event, just call it directly on the singleton

code/datums/observation/observation.dm
```diff
@@ -236,3 +254,46 @@
...
 	return TRUE
+
+/**
+ * Internal proc to handle most all of the observationing procedure
+ *
+ * Will runtime if used on datums with an empty observation list
+ *
+ * Use the [RAISE_EVENT] define instead
+ */
+/datum/proc/RaiseEvent(decl/observ/observation_datum, list/arguments)
```
NanakoAC: delete whole function

code/datums/observation/observation.dm
```diff
@@ -236,3 +254,46 @@
...
+	var/listener = observations[observation_datum]
+	if(!length(listener))
+		return observation_datum.RaiseEvent(listener, src, arguments)
+	. = NONE
+	for(var/I in listener)
+		. |= observation_datum.RaiseEvent(listener, src, arguments)
+
+/decl/observ/RaiseEvent(listener, event_source, list/arguments)
```
NanakoAC: replace this with
```
/decl/observ/RaiseEvent(var/datum/event_source, list/arguments)
	for (var/listener as anything in event_source.observations[src])
		for (var/proctype in event_source.observations[src][listener])
			CallAsync(listener, proctype, arguments)
```

NanakoAC (changes requested):
how to do raise event

code/datums/datum.dm
```diff
@@ -31,6 +38,11 @@
 		if (timer.spent)
 			continue
 		qdel(timer)
+
+	//Unregister need because we will UNREGISTER ALL OBSERVATION_DATUM, NO PROC CALL //TO DO make unregister on OBSERVATION_DATUM
+	for(var/observ in observations)
```
NanakoAC, line 43: wrap this inside `if (observations)` for efficiency

code/datums/observation/observation.dm
```diff
@@ -121,15 +123,25 @@
...
+		event_source.observations = list()
+	if(!event_source.observations[src])
+		event_source.observations[src] = list()
+	if(!event_source.observations[src][listener])
+		event_source.observations[src][listener] =  list()
+
+	// If the proc_call is already registered skip in Singleton
 	if(proc_call in callbacks)
 		return FALSE
+	if(proc_call in event_source.observations[src][listener][event_source])
```
NanakoAC, line 136: this is too many layers deep, delete the `[event_source]`

code/datums/observation/observation.dm
```diff
@@ -121,15 +123,25 @@
...
+		return FALSE
 
 	// Add the callback, and return true.
 	callbacks += proc_call
+	event_source.observations[src][listener][event_source] += proc_call
```
NanakoAC, line 141: same here

code/datums/observation/observation.dm
```diff
@@ -121,15 +123,25 @@
...
 	return TRUE
 
-/decl/observ/proc/unregister(var/event_source, var/datum/listener, var/proc_call)
+/decl/observ/proc/unregister(datum/event_source, datum/listener, proc_call)
 	// Sanity.
 	if (!(event_source && listener && (event_source in event_sources)))
 		return FALSE
```
NanakoAC, line 147: you've not done anything with this, need me to explain how unregister should work?

NanakoAC (changes requested):
unregister  isnt done, plus some small changes

PR #1420, opened 2021-05-13
HOTFIXES

DeadLineSS13:
## This pr solve problem with:
- [ ] OOC chat **(need test)**
- [x]  DOOC chat
- [x] LOOC chat
- [x]  AOOC chat
- [ ]   Solve problem with modes **(need test)**
- [x]   Solve problem with mob health (limbs and etc)

### DIFF
--- a/code/controllers/configuration/entries/event.dm
+++ b/code/controllers/configuration/entries/event.dm
@@ -10,34 +10,28 @@
 /datum/config_entry/flag/ninjas_allowed
 
 /datum/config_entry/flag/abandon_allowed
-    config_entry_value = TRUE
 
 /datum/config_entry/flag/ooc_allowed
-    config_entry_value = TRUE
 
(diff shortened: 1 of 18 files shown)


code/controllers/configuration/entries/event.dm
```diff
@@ -10,34 +10,28 @@
...
-    config_entry_value = TRUE
 
 /datum/config_entry/flag/looc_allowed
-    config_entry_value = TRUE
 
 /datum/config_entry/flag/dooc_allowed
-    config_entry_value = TRUE
 
 /datum/config_entry/flag/dsay_allowed
-    config_entry_value = TRUE
```
NanakoAC, line 25: why are these removed

DeadLineSS13: > why are these removed

in config game_options ready

NanakoAC (approved):

PR #1421, opened 2021-05-14
Rework Planes+layers

DeadLineSS13:
## About this PR
This pr with rework defines plane+layers. Make this more simply and readable. (This thing need to Z-mimic too)
## Founded problems:

- [x] Problems with layer mobs. (FIXED)


NanakoAC (commented):
this tentatively looks good but we need to test it, try things out locally

NanakoAC (approved):
after a bit of testing, i cant find any issues, seems stable enough to merge

PR #1428, opened 2021-05-17
code fixes at infinitum

Evankhell561:
before:
image

after:
image

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -1175,9 +1175,8 @@ datum/admins/var/obj/item/weapon/paper/admin/faxreply // var to hold fax replies
 		P.stamps += "<hr><i>This paper has been stamped by the [P.origin] Quantum Relay.</i>"
 
 		var/image/stampoverlay = image('icons/obj/bureaucracy.dmi')
-		var/{x; y;}
-		x = rand(-2, 0)
-		y = rand(-1, 2)
+		var/x = rand(-2, 0)
+		var/y = rand(-1, 2)
 		P.offset_x += x
 		P.offset_y += y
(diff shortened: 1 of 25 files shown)


code/modules/admin/admin.dm
```diff
@@ -1175,9 +1175,8 @@ datum/admins/var/obj/item/weapon/paper/admin/faxreply // var to hold fax replies
 		P.stamps += "<hr><i>This paper has been stamped by the [P.origin] Quantum Relay.</i>"
 
 		var/image/stampoverlay = image('icons/obj/bureaucracy.dmi')
-		var/{x; y;}
```
NanakoAC, line 1178: this is some wierd syntax, are you sure its not valid?

code/modules/mob/living/carbon/human/species/necromorph/ubermorph.dm
```diff
@@ -277,7 +277,7 @@ Best used near the end, when all seems quiet, to help the necromorphs hunt down
 			//Find target organ should never fail, we won't bother checking
 			targetstring = "[H]'s [found_organ]"
 
-			playsound(charge., 'sound/weapons/slice.ogg', VOLUME_MAX, 1)
```
NanakoAC, line 280: good catch, wrong fix. This sound should be played from charge.victim

charge is a non atomic datum here

NanakoAC (changes requested):
just one fix with the ubermorph

code/modules/admin/admin.dm
Evankhell561: Byond may allow compiling in this format but I try to make everything conform to the standard to which we write code.

PR #1433, opened 2021-05-19
Zmimic

DeadLineSS13:
## Founded problem:
Problem with layers. Need fall if don't there wall.


DeadLineSS13:
TO DO:
- [x] Fix problem with stairs
- [x] Fix problem with layers
- [x] Fix problem with atmospheric

code/__defines/turfs.dm
```diff
@@ -20,4 +20,6 @@
 #define SMOOTH_ALL 1	//Smooth with all of type
 #define SMOOTH_WHITELIST 2	//Smooth with a whitelist of subtypes
 #define SMOOTH_BLACKLIST 3 //Smooth with all but a blacklist of subtypes
-#define SMOOTH_GREYLIST 4 // Use a whitelist and a blacklist at the same time. atom smoothing only
\ No newline at end of file
+#define SMOOTH_GREYLIST 4 // Use a whitelist and a blacklist at the same time. atom smoothing only
+
+#define RANGE_TURFS(CENTER, RADIUS) block(locate(max(CENTER.x-(RADIUS), 1), max(CENTER.y-(RADIUS),1), CENTER.z), locate(min(CENTER.x+(RADIUS), world.maxx), min(CENTER.y+(RADIUS), world.maxy), CENTER.z))
```
NanakoAC, line 25: we already have trange for this, check if this is needed at all

code/controllers/subsystems/ao.dm
```diff
@@ -0,0 +1,31 @@
+SUBSYSTEM_DEF(ao)
+	name = "Ambient Occlusion"
```
NanakoAC, line 2: AO is a native byond feature, why is there a subsystem for it? I don't quite understand the purpose of this

code/controllers/subsystems/lighting.dm
```diff
@@ -50,6 +50,8 @@ SUBSYSTEM_DEF(lighting)
 	for (var/turf/T in (targets || world))
 		if (T.dynamic_lighting && T.loc:dynamic_lighting)
 			T.lighting_build_overlay()
+		// If this isn't here, BYOND will set-background us.
+		CHECK_TICK
```
NanakoAC, line 54: is this the cause of the startup delay?

code/datums/movement/mob.dm
```diff
@@ -2,15 +2,6 @@
 	var/moving           = FALSE
 
 /mob/proc/SelfMove(var/direction)
-	///Mobs with slow turning take a move to turn in place.
-	//We will attempt to turn towards the target if our movement is off cooldown
-	if (slow_turning)
```
NanakoAC, line 7: This is code I added, its necessary for a lot of things, particularly lurkers and brutes. Needs to be preserved

code/datums/movement/movement.dm
```diff
@@ -45,11 +45,3 @@ var/const/MOVEMENT_STOP    = 0x0008
 		if((may_move & (MOVEMENT_PROCEED|MOVEMENT_HANDLED)) == (MOVEMENT_PROCEED|MOVEMENT_HANDLED))
 			return TRUE
 	return TRUE
-
-// Base
-/atom/movable/Destroy()
-	if(LAZYLEN(movement_handlers) && !ispath(movement_handlers[1]))
```
NanakoAC, line 51: why remove this?

code/datums/observation/density_set.dm
```diff
@@ -19,7 +19,7 @@ GLOBAL_DATUM_INIT(density_set_event, /decl/observ/density_set, new)
 *******************/
 /atom/set_density(new_density)
 	var/old_density = density
-	. = ..()
+	UNLINT(. = ..())
```
NanakoAC, line 22: what is unlint?

code/game/machinery/doors/firedoor.dm
```diff
@@ -21,6 +21,7 @@
 	layer = BELOW_DOOR_LAYER
 	open_layer = BELOW_DOOR_LAYER
 	closed_layer = ABOVE_WINDOW_LAYER
+	movable_flags = MOVABLE_FLAG_Z_INTERACT
```
NanakoAC, line 24: this flag is for allowing things to be activated across zlevels? Why put it on firedoors?

code/game/machinery/doors/firedoor.dm
```diff
@@ -442,43 +443,6 @@
 //These are playing merry hell on ZAS.  Sorry fellas :(
 
 /obj/machinery/door/firedoor/border_only
-/*
```
NanakoAC, line 445: I'm not sure why this was all commented out, best leave it as is

code/game/objects/structures/catwalk.dm
```diff
@@ -3,47 +3,25 @@
...
+	var/hatch_open = FALSE
+	var/obj/item/stack/tile/mono/plated_tile
 
 /obj/structure/catwalk/Initialize()
 	. = ..()
-	for(var/obj/structure/catwalk/C in get_turf(src))
-		if(C != src)
-			qdel(C)
-	var/turf/T = get_turf(src)
-	register_zstructure(T)
```
NanakoAC, line 31: this zstructure feature is an addition of mine, to optimise movement. 

Before i added it, things used to check every object in the tile to see if they could prevent falling, now they only check things in the zstructures list

Removing it is fine if you've duplicated the functionality elsewhere, but i don't want to see the bad method returning

code/game/turfs/simulated/footsteps.dm
```diff
@@ -158,4 +158,16 @@
...
+		playsound(T, footsound, volume, 1, range)
+
+/mob/living/carbon/human/proc/has_footsteps()
+	if(species.silent_steps || buckled || lying || throwing)
+		return //people flying, lying down or sitting do not step
+
+	if(shoes && (shoes.item_flags & ITEM_FLAG_SILENT))
+		return // quiet shoes
+
+	if(!has_organ(BP_L_FOOT) && !has_organ(BP_R_FOOT))
```
NanakoAC: Remove this check completely, necromorphs don't have feet

code/game/turfs/space/space.dm
```diff
@@ -40,40 +40,27 @@
...
-
 	if(istype(below, /turf/space))
 		return
 	var/area/A = below.loc
 
 	if(!below.density && (A.area_flags & AREA_FLAG_EXTERNAL))
 		return
 
-	//We found a solid thing below us
-	first_solid_z_below = z-1
```
NanakoAC, line 53: this functionality was added for meteors, it was always a hack, but zmimic needs to replace it

code/game/turfs/turf.dm
```diff
@@ -50,21 +52,39 @@
...
+	. = ..()
 	if(dynamic_lighting)
 		luminosity = 0
 	else
 		luminosity = 1
 
+	if (mapload && permit_ao)
+		queue_ao()
+
+	if (z_flags & ZM_MIMIC_BELOW)
```
NanakoAC, line 65: add this flag to space tiles, i want them to have zmimic functionality too

code/game/turfs/turf.dm
```diff
@@ -303,3 +322,9 @@ var/const/enterloopsanity = 100
 		if(isliving(AM))
 			var/mob/living/M = AM
 			M.turf_collision(src, speed)
+
+/turf/proc/is_open()
```
NanakoAC: I already added a var, is_hole, for this

unless complex functionality is needed, checking that var is probably faster. could at least return it here

code/modules/multiz/movement.dm
```diff
@@ -66,69 +75,45 @@
 /mob/living/carbon/human/can_overcome_gravity()
```
NanakoAC, line 75: add in a check for if the mob is using wallrun here

code/modules/multiz/structures.dm
```diff
@@ -205,28 +200,14 @@
...
-
-
-
-
 /obj/structure/stairs
-	name = "Stairs"
+	name = "stairs"
 	desc = "Stairs leading to another deck.  Not too useful if the gravity goes out."
 	icon = 'icons/obj/stairs.dmi'
-	breakable = FALSE
```
NanakoAC, line 221: stairs being unbreakable was a design decision, put that back in

code/modules/multiz/turf.dm
```diff
@@ -1,90 +1,48 @@
...
-			if (zstructures[B] > highest_priority)
-				var/result=B.CanZPass(A, direction)
-				if (result != ZTRANSITION_MAYBE)
-					highest_priority = zstructures[B]
-					.=result
+			return !density
 
 /turf/simulated/open/CanZPass(atom/A, direction)
-	.=TRUE
-	if (LAZYLEN(zstructures))
```
NanakoAC: this file has the zstructure optimisation I was talking about. this needs to be preserved in some form

NanakoAC (changes requested):
Alright finished, mostly good but a few key concerns:
1. We need this to work for space tiles as well, have them show stuff underneath like openspaces
2. A variety of pre-existing features have been removed, not all should be, see comments

code/__defines/turfs.dm
DeadLineSS13: This check moved to special define turfs

code/controllers/subsystems/ao.dm
DeadLineSS13: > AO is a native byond feature, why is there a subsystem for it? I don't quite understand the purpose of this

Zmimic work with this subsystem

code/controllers/subsystems/lighting.dm
DeadLineSS13: > is this the cause of the startup delay?

No. I ported this because this one of part fixes.

code/datums/movement/movement.dm
DeadLineSS13: > why remove this?

https://github.com/DS-13-Dev-Team/DS13/blob/ff6aaafa5574132b7885cca6cc152f0119788779/code/game/atoms_movable.dm#L59 moved

code/datums/observation/density_set.dm
DeadLineSS13: > what is unlint?

One thing of SpacemanDMM

code/game/machinery/doors/firedoor.dm
DeadLineSS13: > this flag is for allowing things to be activated across zlevels? Why put it on firedoors?

image

code/game/machinery/doors/firedoor.dm
DeadLineSS13: > I'm not sure why this was all commented out, best leave it as is

Old code there. I think this don't will work.

code/game/turfs/simulated/footsteps.dm
DeadLineSS13: > Remove this check completely, necromorphs don't have feet

Add one check if it is no necro

code/game/machinery/doors/firedoor.dm
NanakoAC: right its there, but why?

code/__defines/turfs.dm
NanakoAC: I don't understand this response. Why does this define exist? is trange not good enough?

code/controllers/subsystems/ao.dm
NanakoAC: right but why is it a subsystem? Can you explain what it does?

code/datums/movement/mob.dm
NanakoAC: question not answered

code/game/machinery/doors/firedoor.dm
NanakoAC: it won't, but leave it there for future reference, we will probably want directional firedoors at some point

code/game/turfs/simulated/footsteps.dm
NanakoAC: no, just remove the check. Even if a human has their feet cut off, their leg stumps would still make a sound when walking on them

code/game/turfs/space/space.dm
NanakoAC: what have you done to replace this?

code/__defines/turfs.dm
DeadLineSS13: > I don't understand this response. Why does this define exist? is trange not good enough?

I only moved check. No create new check

code/controllers/subsystems/ao.dm
quardbreak: There's two ways of AO implementation, one based on turf overlays (which is why ao subsystem exists - to control this process) and filter-based AO, which doesn't work with zmimic. If I remember correct, zmimic takes appearance of turf, which contain turf's overlays, so non-filter AO will work with zmimic OK.

code/game/turfs/turf.dm
CrimsonShrike: Space tiles will have undefined behaviour if using zmimic. Correct way to handle it would be to make a new unsimulated turf for vacuum. Space would only be bottom of the map. 

ChangeTurf doesnt support space above things so it's also good for long term correctness

code/datums/movement/mob.dm
DeadLineSS13: > question not answered

already returned.

code/game/machinery/doors/firedoor.dm
DeadLineSS13: > it won't, but leave it there for future reference, we will probably want directional firedoors at some point

returned

code/game/turfs/simulated/footsteps.dm
DeadLineSS13: > no, just remove the check. Even if a human has their feet cut off, their leg stumps would still make a sound when walking on them


The paradox is that here a person steps with his feet and this is clearly audible in the sound. It will be strange that a person without legs will walk and the sounds will be the same.

code/game/turfs/turf.dm
DeadLineSS13: > I already added a var, is_hole, for this
> 
> unless complex functionality is needed, checking that var is probably faster. could at least return it here

Deleted

code/modules/multiz/movement.dm
DeadLineSS13: > add in a check for if the mob is using wallrun here

ready

code/modules/multiz/structures.dm
DeadLineSS13: > stairs being unbreakable was a design decision, put that back in

ready

code/modules/multiz/turf.dm
DeadLineSS13: > this file has the zstructure optimisation I was talking about. this needs to be preserved in some form

ready

code/game/turfs/space/space.dm
DeadLineSS13: > what have you done to replace this?

No. I don't replaced this.

code/game/turfs/simulated/footsteps.dm
NanakoAC: that's a minor edge case which isn't important, just remove the check.

can add in another check for lying if thats important, but i think that might already be checked elsewhere

code/game/turfs/turf.dm
NanakoAC: That's fine to do if you can do it procedurally. Making this work in space is the most important part of this project

it can be done in a second PR though

code/datums/movement/mob.dm
NanakoAC: I don't see it here, it's still missing in the diff?

code/game/turfs/space/space.dm
NanakoAC: Meteors use this var to determine what Zlevel to hit. The thing you've deleted here breaks it. Please look into how it works and fix

NanakoAC (changes requested):
still some comments to deal with

code/datums/movement/mob.dm
DeadLineSS13: > I don't see it here, it's still missing in the diff?

image

code/game/gamemodes/meteor/meteor_object.dm
```diff
@@ -84,14 +84,14 @@ GLOBAL_LIST_EMPTY(asteroids)
 	//That includes empty space with potential non empty space below
 	var/impacting = FALSE
 	if (istype(next_loc))
-		if (istype(next_loc, /turf/space))
+		if (next_loc.z_eventually_space)
```
NanakoAC: this doesn't seem right. 

if z_eventually_space is true, then we shouldn't need to do anymore checks, we can assume the turf is fine to move into ?
it should be true in cases where the turf is space tiles all the way down

NanakoAC:
[meteor_bugs.txt](https://github.com/DS-13-Dev-Team/DS13/files/6663834/meteor_bugs.txt)See attached runtime log. A slew of errors which only occur in this branch

To reproduce: trigger event meteor_wave/ishimura, sit in the asteroid defense cannon, and wait

NanakoAC (changes requested):
the runtime errors with meteors are new and need addressed

NanakoAC (approved):
alright this looks good to go at last

PR #1444, opened 2021-05-27
fix: you can uses kinesis while dead 

### LINKED ISSUE #1143: Kinesis can be used while being unconscious
#### Description of issue:
dunno if this has been reported yet but you can use kinesis while unconcious

This should not be possible.

### LINKED ISSUE #1406: Bug: you can uses kinesis while dead
#### Steps to reproduce: While dying have kinesis on, step two die and use kinesis

https://user-images.githubusercontent.com/54816502/116961561-8269bf80-ac58-11eb-86e1-a5ed279c1bcf.mp4

Evankhell561:
fix: #1406 
fix: #1143
:cl:
fix: Kinesis can be used while being unconscious
/:cl:

### DIFF
--- a/code/_onclick/click_handler.dm
+++ b/code/_onclick/click_handler.dm
@@ -169,10 +169,6 @@ var/const/CLICK_HANDLER_ALL                  = (~0)
 //If its not valid, null will be returned
 //In the case of click catchers, we resolve and return the turf under it
 /datum/click_handler/proc/resolve_world_target(var/a, var/params)
-	if(isliving(user))
-		var/mob/living/L = user
-		if(user.incapacitated())
-			return
 	if (params && user && user.client)
 		var/b = user.client.resolve_drag(a, params)
 		if (a != b)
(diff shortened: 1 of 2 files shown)


code/_onclick/click_handler.dm
```diff
@@ -169,6 +169,10 @@ var/const/CLICK_HANDLER_ALL                  = (~0)
 //If its not valid, null will be returned
 //In the case of click catchers, we resolve and return the turf under it
 /datum/click_handler/proc/resolve_world_target(var/a, var/params)
+	if(isliving(user))
```
NanakoAC: this isn't the place to do this, far too low level

click handlers are generic and used by many things, this problem lies specifically with kinesis

in addition, don't check stat directly, use the incapacitated() proc with the appropriate flag set from defines/mob.dm
https://i.imgur.com/X3GbK9v.png

probably knockout is what you need

Lastly, this is failing a map check, not sure why

NanakoAC (changes requested):

NanakoAC (approved):

PR #1445, opened 2021-05-28
removes duplicate definitions. 

Evankhell561:
before:
before
after:
image


code/__defines/items_clothing.dm
```diff
@@ -101,11 +101,9 @@
 #define slot_r_ear_str      "slot_r_ear"
 #define slot_belt_str       "slot_belt"
 #define slot_shoes_str      "slot_shoes"
-#define slot_head_str      	"slot_head"
```
NanakoAC, line 104: i added these, don't remove them

code/__defines/misc.dm
```diff
@@ -191,7 +191,6 @@
 #define CELLS 8								//Amount of cells per row/column in grid
 #define CELLSIZE (world.icon_size/CELLS)	//Size of a cell in pixels
 
-#define WORLD_ICON_SIZE 32
```
NanakoAC, line 194: this is very necessary, is it duplicated elsewhere?

code/__defines/subsystems.dm
```diff
@@ -1,22 +1,7 @@
-//Timing subsystem
-//Don't run if there is an identical unique timer active
-//if the arguments to addtimer are the same as an existing timer, it doesn't create a new timer, and returns the id of the existing timer
-#define TIMER_UNIQUE		0x1
-//For unique timers: Replace the old timer rather then not start this one
-#define TIMER_OVERRIDE		0x2
```
NanakoAC, line 6: this seems like the place for these to be

NanakoAC (changes requested):
some of these im not sure the correct duplicate has been removed

code/__defines/items_clothing.dm
Evankhell561: Higher up in this dm is this same defined repeated.
image

code/__defines/misc.dm
Evankhell561: image

code/__defines/subsystems.dm
Evankhell561: You tell me which of these two files you prefer to keep these defines.
image

NanakoAC: subsystems.dm

NanakoAC (approved):

PR #1448, opened 2021-05-29
removes some code degug messages.

Evankhell561:
:cl:
code: removes some code degug messages.
/:cl:

### DIFF
--- a/code/modules/client/patron.dm
+++ b/code/modules/client/patron.dm
@@ -28,6 +28,7 @@
 		if (line_list.len > 1)
 			date = sanitize_date(line_list[2])	//This will return FALSE if the date does not look right
 		if (!date)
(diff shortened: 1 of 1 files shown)


code/modules/client/patron.dm
```diff
@@ -25,11 +25,9 @@
 		var/key = ckey(line_list[1])
 
 		var/date
-		log_world("Doing key [key], date is [date]")
 		if (line_list.len > 1)
 			date = sanitize_date(line_list[2])	//This will return FALSE if the date does not look right
 		if (!date)
-			log_world("ERROR: Bad date found in patrons.txt for ckey [key], replacing this with a 1 month timer from now as a fallback")
```
NanakoAC: this one is fine to keep in live code, it actually indicates something bad which shouldn't happen

NanakoAC (changes requested):

NanakoAC (approved):

PR #1452, opened 2021-05-31
Replaced IC's with newer

### LINKED ISSUE #1446: Suggestion: Replace current integrated circuits with baystation variant
#### Suggestion:
Replace current integrated circuits with baystation variant.

It'll add more things to do for RnD. The applications for integrated circuit boards are endless. You can make some simple things like speaking box, slow portable recharges that don't need much place or you can spend half an hour trying to make automatic medical scaner + healer. You also can make flashlight's that will follow you and won't take place in your inventory.
I don't think that it is OP because not all players know how to use them/have premade circuits + you need to spend enough steel to make actually cool machines.

DTraitor:
### Purpose
This PR was created to replace old integrated circuits with their new variant. Now in order to make them RnD has to do some simple researches and produce "Integrated Circuit Printer". It can be upgraded with "Integrated Circuit Printer Upgrade Disk" which allows you to produce some parts that were previously blocked. I've also removed few modules because I think that they are useless. These are slime scanner, plant scanner, plant manipulator, gene scanner (for seeds) and bluespace rift generator (OP and useless at the same moment). Augmentaion assemblie was also removed because it requires additional changes to the way prothestic parts work. I was too lazy to understand how new config system works so in order to disable them you have to edit "configuration.dm". There is still button in admin's secrets to disable IC's printing.

### Notes
The only reason I decided to do this PR is integrated circuit's printer in RnD that was useless since you couldn't produce tools to work with IC's. Printer was replaced by a workbench since they look similar and it's useful to have one in the place where necromorphs cant be at round start.
Closes: #1446

I don't know why there is so much edited lines in Ishimura since I just replaced one thing with another.


Snypehunter007:
@DTraitor can you give an overview of the differences between the two versions for those of us who don't often or at all play Research?

What are the exact uses of the Gene Scanner and Plant Manipulator? I only ask because on the surface level they sound needed for Botany. 

What do you mean by "so many edited lines lines in the Ishimura" btw?

Lastly, thank you very much for following the .yml format. It is the small things that really help.

DTraitor:
>can you give an overview of the differences between the two versions for those of us who don't often or at all play Research?

The biggest difference between them is amount of assemblies (5 types and their subtypes that look different but have the same stats, ability to print pre made circuits after you have upgraded your printer and now IC's need power to work. Also there is a lot of new modules (grabber, pulling claw, thrower, anchoring bolts, securitron sound circuit, access circuit etc). In the version we have you cant make circuit wirer and other tools you need to make assemblies so there is no reasons to have printer in RnD at all. Plus some circuit's cost too much steel.
>What are the exact uses of the Gene Scanner and Plant Manipulator?

They are used for botany when you need to get plant stats or gather fruits but we loose botany almost immediately after marker awake. In case you'll change botany position in Ishimura rework I can add them back.
> What do you mean by "so many edited lines lines in the Ishimura" btw?

I've replaced only one thing and github says I've removed 82,000 of lines

NanakoAC:
> They are used for botany when you need to get plant stats or gather fruits but we loose botany almost immediately after marker awake. In case you'll change botany position in Ishimura rework I can add them back.

Being unbalanced is not a reason for deleting content, put those back

NanakoAC:
> > What do you mean by "so many edited lines lines in the Ishimura" btw?
> 
> I've replaced only one thing and github says I've removed 82,000 of lines

you have, this is a 90 file PR, it is huge. I hope you know what you're doing here, what kind of experience do you have doing this kind of stuff?

I don't mean to discourage, we always welcome new contributors. But this is quite ambitious for a first submission, it's going to take a while to review. Are you confident in the code integrity? Will you be able to fix issues as they're highlighted?

DTraitor:
>Being unbalanced is not a reason for deleting content, put those back

Done

NanakoAC (approved):
Alright after giving this a very thorough examination, I have a few things to say;

first of all, this is pretty huge. but it's fairly selfcontained which is a major mitigating factor to the risk and complexity.

This will, undoubtedly, cause some balance issues, but they're going to be pretty fun stories in the process. I don't see major potential for breaking or ruining the game, and this seems like it'll be a net benefit overall

DTraitor:
Anchoring bolts had no icon

NanakoAC (changes requested):
just got a few things to fix in the workflows

NanakoAC (approved):

PR #1460, opened 2021-06-03
Scorched earth

NanakoAC:
changes: 
  - rscadd: "When corruption is destroyed via heat effects (fire, lasers, explosions, etc) the scorched floor will block corruption regrowth for a few minutes afterwards"
  - bugfix: "Made sure corruption can now go through deployable barricades and knocked-down-vendors"
  - bugfix: "Deployable barriers can now be attacked by necromorphs"

Also a little additional fix, edits default config to allow admin jumping again. that got caught up accidentally in the recent config rework


Ketrai:
This feels like the kind of feature that will need a dedicated test round to pluck out potential lag issues or bugs.

PR #1465, opened 2021-06-07
Overlays system

DeadLineSS13:
## About this PR
Overlays subsystem.


code/game/atoms.dm
```diff
@@ -31,7 +31,12 @@
 	var/default_alpha = 255
 	var/default_scale = 1
 
-
+	///overlays that should remain on top and not normally removed when using cut_overlay functions, like c4.
+	var/list/priority_overlays
+	///a very temporary list of overlays to remove
+	var/list/remove_overlays
+	///a very temporary list of overlays to add
+	var/list/add_overlays
```
NanakoAC, line 39: This is not a good idea, adding three new lists to every atom in the game.

How much of this is ported? I seem to recall the PRs you showed me only added one list, not three

Couldn't this data be stored in an extension for the things that need it, instead of new vars?

NanakoAC (changes requested):

code/game/atoms.dm
DeadLineSS13: > This is not a good idea, adding three new lists to every atom in the game.
> 
> 
> 
> How much of this is ported? I seem to recall the PRs you showed me only added one list, not three
> 
> 
> 
> Couldn't this data be stored in an extension for the things that need it, instead of new vars?

https://github.com/tgstation/tgstation/blob/46acdbd52d98ee79fb4b823b5c09676756c0d8e4/code/game/atoms.dm#L52

PR #1466, opened 2021-06-07
Infinite plasma pickaxe and plasma saw fix

DTraitor:
### Purpose

Fixes infinite plasma pickaxe and plasma saw. 
Right now you can turn on plasma pickaxe, remove it's battery and have infinite plasma pickaxe. This PR fixes it.

### DIFF
--- a/code/game/objects/items/weapons/tools/_tools.dm
+++ b/code/game/objects/items/weapons/tools/_tools.dm
@@ -311,6 +311,8 @@
 //Cell reload
 /obj/item/weapon/tool/MouseDrop(over_object)
 	if((src.loc == usr) && istype(over_object, /obj/screen) && eject_item(cell, usr)) ///Formerly checked for /obj/screeninventory/hand. Not sure how that fits with bay's UI system
+		if (switched_on)
+			turn_off()
 		cell = null
 		update_icon()
 	else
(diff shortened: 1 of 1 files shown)


code/game/objects/items/weapons/tools/_tools.dm
```diff
@@ -311,6 +311,8 @@
 //Cell reload
 /obj/item/weapon/tool/MouseDrop(over_object)
 	if((src.loc == usr) && istype(over_object, /obj/screen) && eject_item(cell, usr)) ///Formerly checked for /obj/screeninventory/hand. Not sure how that fits with bay's UI system
+		if (switched_on)
```
NanakoAC, line 314: this doesn't seem the right place to do this

Anything with a power cost should be consuming power each tick, and turning off if it fails to do so. Find out why that isn't working

NanakoAC (changes requested):

code/game/objects/items/weapons/tools/_tools.dm
DTraitor: I've just checked few things and I've discovered that tools don't drain power passive. It drains power on use and when someone turns it on.

DTraitor:
Tools will now turn off when you try to use them and there is no power left

code/game/objects/items/weapons/tools/_tools.dm
NanakoAC: passive power drain is definitely a thing, it must be broken., fixing that is the way to resolve this problem

NanakoAC:
Dtraitor asked me to close this

PR #1468, opened 2021-06-08
Necromorphs will no longer speak english

### LINKED ISSUE #1336: Bug: Necros can speak english
#### Description of issue:
Necros can speak galatic common

Necros should not know common

#### Steps to reproduce:
Go to langauge tab and select galatic common as a necro

DTraitor:
### Purpose
Necromorphs will no longer speak english
Fixes: #1336


code/modules/mob/living/carbon/human/species/necromorph/necromorph_species.dm
```diff
@@ -125,7 +125,8 @@
 	species_flags = SPECIES_FLAG_NO_PAIN | SPECIES_FLAG_NO_MINOR_CUT | SPECIES_FLAG_NO_POISON  | SPECIES_FLAG_NO_BLOCK      // Various specific features.
 	appearance_flags = 0      // Appearance/display related features.
 	spawn_flags = SPECIES_IS_RESTRICTED | SPECIES_NO_FBP_CONSTRUCTION | SPECIES_NO_FBP_CHARGEN           // Flags that specify who can spawn as this specie
-	language = LANGUAGE_NECROCHAT
+	language = null
+	default_language = "Signal Hivemind"
```
NanakoAC, line 129: don't use string literals when a define exists, stick with that

why not set it in both vars too?

NanakoAC (changes requested):

code/modules/mob/living/carbon/human/species/necromorph/necromorph_species.dm
DTraitor: **#define** doesnt work in this case and the game can't find language to use so it automatically sets it to english common. Var is declared in **species.dm**.

_P.S. Monkeys use the same code as I put here_

DTraitor:
whoops, merged to wrong branch

code/modules/mob/living/carbon/human/species/necromorph/necromorph_species.dm
NanakoAC: well that definitely doesn't make sense. Perhaps the define is mis-spelled, try looking into it

DTraitor:
Closing this because I couldn't find a way to use **#define** in this case (it doesnt work because it requires var in **say.dm**), **var/language** is declared in **species.dm** and changing it using simple "language = null "doesnt arranges @NanakoAC.

PR #1496, opened 2021-06-11
Stamina system

DTraitor:
### Purpose
Ports stamina system, doing some balance to move speed and adds third move intent.
Full list of changes:
- Added stalk intent which makes no noise but it's quiet slow;
- Added baystation stamina system, walk is a main move intent now and has almost the same speed as run before;
- In order to sprint you can hold shift or change your move intent to run.
- Necromorphs are unable to run. They should use charge in order to catch people (probably some configuration needed)
- Lying and eating (having nutritions in your stomach) makes restoring stamina faster.
- Stamina consumption depends on athletics skill and your gear
- Stamina system works different in zero gravity.

Original PR: https://github.com/Baystation12/Baystation12/pull/25773


code/__defines/movement.dm
```diff
@@ -30,4 +30,12 @@
 #define S_SOUTHEAST	"6"
 #define S_SOUTHWEST	"10"
 #define S_UP	"16"
-#define S_DOWN "32"
\ No newline at end of file
+#define S_DOWN "32"
+
+#define run_delay 2
```
NanakoAC: defines should use allcaps to distinguish them from normal variables

code/_onclick/hud/human.dm
```diff
@@ -34,14 +34,17 @@
 	if(hud_data.icon)
 		ui_style = hud_data.icon
 
-	src.adding = list()
-	src.other = list()
+	adding = list()
```
NanakoAC, line 37: any particular reason for this change?

src. is usually unnecessary but sometimes it can be useful, as a signal to the compiler or the programmer, which context is meant

code/datums/move_intent/move_intent.dm
```diff
@@ -8,7 +8,20 @@
 	var/flags = 0
 	var/move_delay = 1
 	var/hud_icon_state
-	var/footstep_interval = 1
+
+/decl/move_intent/proc/can_be_used_by(var/mob/user)
```
NanakoAC, line 14: when is this check run?

We need to prevent people from sprinting when they're not standing up, this may be a good place for that deprending on when its called

If, as it seems, this is just used to prevent necromorphs from sprinting, it'd be better not to have this intent in their list at all

code/__defines/movement.dm
DTraitor: done

code/game/turfs/simulated/footsteps.dm
```diff
@@ -125,6 +125,9 @@
 	if(step_count % step_interval)
 		return
 
+	if(istype(move_intent, /decl/move_intent/stalk)) //We don't make sounds if we're tiptoeing
```
NanakoAC: this is sloppy, add a flag to move intents indicating that they don't make sounds instead,, or specifying a really long sound interval

code/modules/mob/living/carbon/human/human_movement.dm
```diff
@@ -62,7 +66,8 @@
 			else if(E.status & ORGAN_BROKEN)
 				tally += 1.5
 
-	if(shock_stage >= 10) tally += 3
+	if(shock_stage >= 10 || get_stamina() <= 0)
```
NanakoAC: can stamina actually go below zero?

change this value to something useful, like under 30

code/modules/mob/living/carbon/human/life.dm
```diff
@@ -84,6 +87,27 @@
...
+			to_chat(src, SPAN_WARNING("You are exhausted!"))
+			if(MOVING_QUICKLY(src))
+				set_moving_slowly()
+	if(last_stamina != stamina && hud_used)
+		hud_used.update_stamina()
+
+/mob/living/carbon/human/proc/handle_stamina()
+	if((world.time - last_quick_move_time) > 5 SECONDS)
+		var/mod = (lying + (nutrition / initial(nutrition))) / 2
+		adjust_stamina(max(minimum_stamina_recovery, maximum_stamina_recovery * mod) * (1+chem_effects[CE_ENERGETIC]))
```
NanakoAC: reduce the minimum stamina gain a bit to make this calcultion more meaningful

code/modules/mob/living/carbon/human/human_movement.dm
NanakoAC: also can you use the direct number rather than calling a proc here? this proc is called with every single step a human takes, optimising is important

code/datums/move_intent/move_intent.dm
DTraitor: > it'd be better not to have this intent in their list at all

This is how it currently works.
**can_be_used_by** was in that PR so I moved it without changes.

NanakoAC (changes requested):
overall, mostly good, a few tweaks

code/modules/mob/living/carbon/human/human_movement.dm
DTraitor: > can stamina actually go below zero?

There is nothing preventing stamina going below 0 except the fact that you cannot run while you have 0 stamina and there is no other way to use stamina rather than running.
> change this value to something useful, like under 30

This is used to make people move slower when they've spent all their stamina.

NanakoAC (approved):

PR #1502, opened 2021-06-16
Emergency RIG rebalancing

NanakoAC:
Balance feedback about this new RIG was not addressed during the review process.
This PR Addresses those issues.

There are three changes:
-Emergency rigs now have significant slowdown
-They now have minor armor values
-They are much less common, not in every locker. Chance to spawn is now 25%

There are a few reasons for these changes:

1. Superiority.
 In its original implementation, the Emergency rig is clearly superior to the civilian rig. Just as easily obtainable, with the same armor and movement (ie, none), but it comes with the addition of a sealed suit which protects against acid and vacuum, and built in thrusters.
A straight upgrade like this shouldn't be so easily obtainable. 

2. Ubiquity. 
Everyone in the dead space universe wears a rig, but most of those wear the civilian rig, a low profile frame with no suit attached. Rigs with space-capable suits are still uncommon, and its important to ensure any new rig doesn't change that and incentivise everyone to walk around in a space suit all the time

3. Purpose
As the name implies an emergency rig is for emergencies, not wearing all-round. Providing tons of benefits with no drawbacks gives zero incentive to take it off

4. Underappreciation of speed
Movespeed is important. Lower speed penalties are a defining factor of things like the Advanced RIG, which combines being lightweight with being spaceproof. It doesn't make sense for something common to be even better at that. The emergency rig should be slower and bulkier than other spaceproof suits

5. Armor
Slightly contradictory to the above, but a matter of immersion. Wearing a half inch of rubber and plastic on your body should provide some protection for the sake of realism. More to the point, it very definitely looks armored, with clearly defined shoulderpads. This is a matter of meeting player expectations, this thing being unarmored is counterintuitive. Armor values have been factored into the speed penalty, and it is still inferior to almost every other spaceproof suit, having approximately half the defensive values of the mining rig (which was our previous bottom-tier suit)

6. Performance
A relatively minor concern, but not to be overlooked. RIGs are complex items in terms of memory footprint, they contain many sub-datums. Flooding the map with hundreds of extra unused rigs is not a sensible action from a technical perspective. 

With these points in mind, the emergency rig has been rebalanced to better fit its purpose. It is an undesireable item of last resort, definitely not anyone's first choice, and something you're going to want to take off as soon as the emergency has passed. It combines bulky, slow movement, with relatively low armor values and a dearth of modules. But if you desperately need a spaceproof suit, it is still readily available.


Snypehunter007:
You missed literally every reason I said that the armor had the values it had and was in what it is.

This was already discussed and finished with.

ghost:
The balancing of RIGs, weapons and Necromorphs is my own current endeavor right now, and I won't commit to these fixes until I run my own tests on it.

NanakoAC:
> This was already discussed and finished with.

If any discussion was had about this, I wasn't invited. my concerns on the PR were dismissed and ignored

I have highlighted reasons here why it needs rebalanced. The chiefest among them being its superiority over the civilian rig in every way, that can't be allowed to exist in a ubiquitous item

PR #1503, opened 2021-06-16
Faster bioprothestic 

DTraitor:
### Purpose

Growing a limb or an organ will now be twice faster. Makes bioprothestic more useful.
~6 minutes to grow arm or leg
~3 minutes to grow any organ

### DIFF
--- a/code/game/machinery/ds13/bpl/recycling_tank.dm
+++ b/code/game/machinery/ds13/bpl/recycling_tank.dm
@@ -13,7 +13,7 @@
 	var/list/content_atoms = list()
 
 
-	var/breakdown_rate = 0.001	//Remove this many units of biomass per tick, and convert it into purified biomass
+	var/breakdown_rate = 0.03	//Remove this many units of biomass per tick, and convert it into purified biomass
 	var/reagent_breakdown_rate = 0.065	//Remove this many units of reagents per tick and convert to biomass
 
 	var/obj/structure/reagent_dispensers/biomass/storage
--- a/code/game/machinery/ds13/bpl/recycling_tank.dm
+++ b/code/game/machinery/ds13/bpl/recycling_tank.dm
@@ -13,7 +13,7 @@
 	var/list/content_atoms = list()
 
 
-	var/breakdown_rate = 0.03	//Remove this many units of biomass per tick, and convert it into purified biomass
+	var/breakdown_rate = 0.002	//Remove this many units of biomass per tick, and convert it into purified biomass
 	var/reagent_breakdown_rate = 0.065	//Remove this many units of reagents per tick and convert to biomass
 
 	var/obj/structure/reagent_dispensers/biomass/storage
--- a/code/modules/organs/external/_external.dm
+++ b/code/modules/organs/external/_external.dm
@@ -65,7 +65,7 @@
 	var/list/implants = list()         // Currently implanted objects.
 	var/base_miss_chance = 0          // Chance of missing.
 	var/genetic_degradation = 0
-	biomass = 2	//By default, external organs are worth 2kg of biomass. Hella inaccurate, could find more exact values
+	biomass = 1	//How much biomass worth external organs
 
 
 	//Forensics stuff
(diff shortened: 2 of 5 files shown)


DTraitor:
I think it's too fast but let's try twice faster

NanakoAC (changes requested):
I'm afraid this is the wrong way to do this. The biomass of organs has meaning for a lot of things, notably for how much necromorphs get when absorbing severed bodyparts.

To speed up organ growth in the BPL, those machines need to be modified

Now that said, doubling the speed would also be too much for a single jump. We use an iterative balance policy. Changes to values should be 10-25% at a time, then see how things play out.

Lastly, a more general overlook: The BPL is not meant to be fast. You are meant to grow a catalogue of organs in advance and maintain them throughout the round for when they're needed. This is something for medical to take care of. If you start growing a new arm just after someone gets theirs cut off, you're doing it wrong

ghost:
> Lastly, a more general overlook: The BPL is not meant to be fast. You are meant to grow a catalogue of organs in advance and maintain them throughout the round for when they're needed. This is something for medical to take care of. If you start growing a new arm just after someone gets theirs cut off, you're doing it wrong

People are no longer using the BPL because of faster alternatives. It would benefit us to up the speed so it doesn't become redundant.

DTraitor:
> To speed up organ growth in the BPL, those machines need to be modified

If I've edited efficency it would still require a lot of biomass and it takes time to make it. Or I could change formula a bit and that would probably cause some problems like infinite biomass or something like that.

NanakoAC:
alright here's what to do, revert these changes first

On the recycling tanks, double the breakdown rate vars
on the growth tanks, double the growth rate var

and remove the code in growth tank process which applies sustain rate, lets just remove sustain cost completely, it doesn't fit the intended playstyle

Don't worry about quantities of biomass, that's what the nutricubes are for, they provide enough mass for 16 organs, just a matter of breaking it down fast enough

DTraitor:
Recycling food is now much faster, blood is still the fastest way to obtain biomass because it's a reagent. Time to make limb or organ is the same as they were in the first commit.

code/game/machinery/ds13/bpl/recycling_tank.dm
```diff
@@ -13,7 +13,7 @@
 	var/list/content_atoms = list()
 
 
-	var/breakdown_rate = 0.001	//Remove this many units of biomass per tick, and convert it into purified biomass
```
NanakoAC, line 16: I said double, you've increased it by 3000%

target value here is 0.002

code/modules/organs/external/_external.dm
```diff
@@ -65,7 +65,7 @@
 	var/list/implants = list()         // Currently implanted objects.
 	var/base_miss_chance = 0          // Chance of missing.
 	var/genetic_degradation = 0
-	biomass = 2	//By default, external organs are worth 2kg of biomass. Hella inaccurate, could find more exact values
+	biomass = 2	//By default, external organs are worth 2kg of biomass. Hella inaccurate, could find more exact valu
```
NanakoAC: you added a typo here

code/game/machinery/ds13/bpl/recycling_tank.dm
```diff
@@ -13,7 +13,7 @@
...
+	var/breakdown_rate = 0.03	//Remove this many units of biomass per tick, and convert it into purified biomass
 	var/reagent_breakdown_rate = 0.065	//Remove this many units of reagents per tick and convert to biomass
```
NanakoAC, line 17: increase this one too

NanakoAC (changes requested):
the breakdown rate has been increased too much

code/game/machinery/ds13/bpl/recycling_tank.dm
DTraitor: I tried 0.003 and it gives 10 biomass per minute which means that we can make 1 limb in 10 minutes

code/game/machinery/ds13/bpl/recycling_tank.dm
DTraitor: I think 0.065 is enough. It'll turn reagents into biomass pretty fast

NanakoAC (approved):

PR #1507, opened 2021-06-17
Stores and Credits, Phase 1

NanakoAC:
This PR is the first phase of the stores and credits system. Here's what's included:

-The store kiosk, a new machine that allows a crewmember to purchase new gear
-The transfer feature, doing a makeover animation to put the user into a new rig
-The store kiosk comes with a HUGE catalogue, as it hooks into the shared design system. (some things may still be missing) The store catalogue is currently fully unlocked
-Store prices are taken from the games where possible, rough ballpark figures have been used where there's no reference. Despite this, store prices are not really balanced, a lot of things are going to feel too cheap or too expensive, there will be a lot of pricing changes based on player feedback
-A significant reworking of accounts and money. Bank accounts are now withdrawal-only, depositing is not allowed
-Each RIG now contains its own credit account to which money can be transferred in and out at a store, or by using a credit chip
-Wages are now paid out every 30 minutes to employee accounts, and can be withdrawn at stores
-Credit chips now appear in random loot, containing randomised quantities. rarity scales with value
-Credits are currently NOT persistent yet. But the initial account balance has been left in to compensate, so people are starting with some free money every round. This will be removed with persistence.
-A variety of store machines have been mapped into the ship. Their positions have been carefully chosen to serve a variety of purposes. 
	--In some places it's convenience, like the ones in equipment rooms. 
	--In some places (diner notably) it's foot traffic. There are two stores in there to accomodate a larger number of people, and to help promote the diner as a social location
	--Stores are also placed on tram platforms, this is convenient and also canon
	--One has also been placed in the rest stop room on the EVA deck, to serve teams preparing for a run on the asteroid cannon
	
Not every position is set in stone, but significant thought has been put into these, and adding or removing machines is something which shouldn't be taken lightly

Here's what's not yet included, and will come in phase 2
-The store catalogue will be unlocked one item at a time through persistent schematics
-Credits will be made persistent and subject to fees between rounds depending on your survival status
-Store prices will certainly be changed based on feedback
-Some items are definitely missing from the store catalogue
-The transfer feature doesn't currently move accounts between rigs, will fix that

changes: 
  - rscadd: "Added Store kiosks, where players can go purchase things with credits."
  - rscadd: "Store kiosks include a cool transfer feature, which can switch you to a new RIG, and install modules for you, quickly and stylishly."
  - tweak: "Reduced the damage of Divet and Plasma Cutter by 20%. They are meant to be sidearms and not as powerful as main weaponry."
  - tweak: "Reduced the damage of HV Pulse ammo, from 50% to 40% more than normal pulse rounds. A tradeoff for being more readily available through the store."


ghost (approved):
Everything looks good. Two notes; we will need to obviously adjust prices and remove some items after the results of the testing come through. We will also need to lower the loot spawns for high value items exponentially now that Stores are the main avenue.

PR #1508, opened 2021-06-18
Adds an uplink for the EarthGov Agent and for the Unitologist Operative

Evankhell561:
:cl:
add: EarthGov Agent uplink
add: Unitologist Operative uplink
/:cl:


NanakoAC (changes requested):
something to add here;

Uplinks on other servers typically come with a cash briefcase option, not sure if that's here, but if it is remove it and instead replace with preloaded credit chips. check credits.dm, the ewallet

ghost:
I've given Evankhell a list of items for both the EG and Uni Uplinks, they're predetermined and are more focused on gear. Credits are not involved.

NanakoAC (approved):
alright then

PR #1518, opened 2021-06-27
Adds in a silenced divet, and new ammo types

Kmc2000:
image

At west's behest

Will require a testmerge beforehand to check for balance!


code/modules/projectiles/guns/ds13/divet.dm
```diff
@@ -26,6 +30,11 @@
...
 	magazine_type = null
 
+/obj/item/weapon/gun/projectile/divet/silenced
+	icon_state = "divet_spec"
+	name = "special ops divet pistol"
+	silenced = TRUE
+
 /obj/item/weapon/gun/projectile/divet/update_icon()
 	..()
 	if(ammo_magazine && ammo_magazine.stored_ammo.len)
```
Kmc2000, line 40: This was here before but, I feel like this may be a logic error. It's telling the divet to become "divet_e" when there's a loaded mag with ammo in it?

Haven't touched this as I'm uncertain

NanakoAC: look at the sprites and make that determination

code/modules/projectiles/guns/ds13/divet.dm
```diff
@@ -60,3 +103,28 @@
 	armor_penetration = 5
 	structure_damage_factor = 1.5
 	penetration_modifier = 1.1
+	icon_state = "divet"
+
+/obj/item/projectile/bullet/ls_slug/hollow_point
```
NanakoAC, line 108: these should have higher damage and embed chance compared to normal rounds
go with 25% more damage, and 4x embed chance

code/modules/projectiles/guns/ds13/divet.dm
```diff
@@ -60,3 +103,28 @@
...
+	structure_damage_factor = 0.5
+	penetration_modifier = 0
+	armor_penetration = -50
+	icon_state = "divet_hp"
+
+/obj/item/projectile/bullet/ls_slug/ap
+	structure_damage_factor = 1.75
+	penetration_modifier = 1.5
+	armor_penetration = 25 //May wanna change this!
```
NanakoAC: make this 15

PR #1522, opened 2021-06-28
Player constructed machines fix

DTraitor:
### Purpose
Half-fixes machine construction. Player constructed machines now work as inteded and cant be passed by. Still there are some objects that were not fixed because they have additional parameters in New(). They are listed in issue #1524.  Most of them can't be constructed by player right now (except fire alarm, air alarm and light fixture).


code/game/machinery/atmoalter/canister.dm
```diff
@@ -366,72 +366,72 @@ update_flag
...
 	return 1
 
 //Dirty way to fill room with gas. However it is a bit easier to do than creating some floor/engine/n2o -rastaf0
-/obj/machinery/portable_atmospherics/canister/sleeping_agent/roomfiller/New()
+/obj/machinery/portable_atmospherics/canister/sleeping_agent/roomfiller/New(var/atom/location, var/direction, var/nocircuit = FALSE)
 	..()
 	air_contents.gas["sleeping_agent"] = 9*4000
 	spawn(10)
-		var/turf/simulated/location = src.loc
+		var/turf/simulated/location2 = src.loc
```
NanakoAC, line 407: whats this change for?

NanakoAC (approved):

code/game/machinery/atmoalter/canister.dm
DTraitor: because of this:
`code\game\machinery\atmoalter\canister.dm:407:error: location: duplicate definition
code\game\machinery\atmoalter\canister.dm:403:error: location: previous definition`
and if you try to remove` var/turf/simulated/location2 = src.loc` leaving `location.air` and `location.assume_air` it will complain about undefiend var:
`location.air: undefined var`

I forgot to change second `location`, pushed commit that fixes it already

PR #1536, opened 2021-07-09
Telecrystal balance 2

NanakoAC:
changes: 
  - tweak: "Reduced uplink telecrystal quantity from 130 to 30"


Ketrai:
You can still get 3 shrapnel grenades per person like this. Which is potentially 9 in total. We don't need unis to get lazy an pick literally the easiest option ever for guaranteeing a necro minor in normal play.

PR #1551, opened 2021-07-29
Functional Voiceboxes (or "hear your screams, and make them too")

KeRSedChaplain:
Adds in sound files for screaming, gasping, coughing, laughing, and being in pain, split between each gender. Courtesy of DTraitor for making the emotes actually use more than the same sound each time.


NanakoAC (commented):
This seems unfinished, these were meant to be linked to getting hurt. I mentioned we already have functionality for it, override play_species_audio for humans and check for gender

PR #1557, opened 2021-08-03
Tail target

DTraitor:
### Purpose
Added ability to target tails. Cycle groin to do so.


NanakoAC (approved):

NanakoAC:
while fine to have, this is no substitute for actually fixing the bug with limb targeting on guns


firing at the legs on a leaper should hit its tail

PR #1570, opened 2021-08-16
SSoverlays, attemp 2

DTraitor:
### Purpose
Ports TG overlays subsystem with all improvments. According to original PR it's more optimised compared to `overlays +-`. Subsystem applies all overlays in one moment (when fires, it doesn't fire if there is no overlays to change).
> Added a simple generic benchmark stat tracking system.
>
> Overlays will now log how long each type took to process.

 - MrStonedOne, PR 32371

Also moved all overlay changes to related procs.

**Main PR's:**
https://github.com/tgstation/tgstation/pull/23922
https://github.com/tgstation/tgstation/pull/32371
https://github.com/tgstation/tgstation/pull/38002 (information about performance is here)


code/__defines/subsystems.dm
```diff
@@ -96,3 +97,22 @@
...
 #define MACHINE_PROCESS_INTERVAL	2 SECONDS
 #define EVENT_PROCESS_INTERVAL	2 SECONDS
+
+//! ## Overlays subsystem
+
+///Compile all the overlays for an atom from the cache lists
+// |= on overlays is not actually guaranteed to not add same appearances but we're optimistically using it anyway.
+#define COMPILE_OVERLAYS(A) \
+	do {\
+		var/list/ad = A.add_overlays;\
```
NanakoAC: Why are these list vars created? If the add/remove overlays vars already exist, just use them and don;t call cut

code/modules/admin/verbs/debug.dm
```diff
@@ -539,6 +539,13 @@
 	cmd_analyse_health(H)
 	feedback_add_details("admin_verb","ANLS") //If you are copy-pasting this, ensure the 2nd parameter is unique to the new proc!
 
+/client/proc/cmd_display_overlay_log()
```
NanakoAC: this kind of logging seems unnecessary for production code, and is likely to negate some of the performance benefit.

I can't see a compelling reason to leave this in for normal gameplay

NanakoAC (changes requested):
whew that was huge

just the debug verb i think

NanakoAC (approved):

PR #1573, opened 2021-08-18
A lot of SStimer changes

### LINKED ISSUE #1125: Uncommon instance where abilities go into negative cooldown states with no recovery
Description of issue: When I was a normal Slasher, I used the charge shortcut against a guy with a polyacid spray bottle. As soon as I clicked charge, the charge entered a permanent cooldown where I wasn't allowed to charge as it said I was already charging. This did the same to my dodge.

Difference between expected and actual behavior: I should've been able to execute the charge but again it said I was already charging when I had been moving around like normal instead of the dashing movement of a charge.

Steps to reproduce: First, use the charge shortcut and then be struck by polyacid during a lagspike.

DTraitor:
### Purpose
Ports a lot changes and fixes to SStimer from tgstation.
Remaining time now can be checked using `timeleft(id)`.
Following PR's were used (not all changes were ported):
 * https://github.com/tgstation/tgstation/pull/33098
 * https://github.com/tgstation/tgstation/pull/34208
 * https://github.com/tgstation/tgstation/pull/38008
 * https://github.com/tgstation/tgstation/pull/38009
 * https://github.com/tgstation/tgstation/pull/38790
 * https://github.com/tgstation/tgstation/pull/38994
 * https://github.com/tgstation/tgstation/pull/38995
 * https://github.com/tgstation/tgstation/pull/39747 **|| Fixes: https://github.com/DS-13-Dev-Team/DS13/issues/1125**
.
 * https://github.com/tgstation/tgstation/pull/40417
 * https://github.com/tgstation/tgstation/pull/40947
 * https://github.com/tgstation/tgstation/pull/43003
 * https://github.com/tgstation/tgstation/pull/46447
 * https://github.com/tgstation/tgstation/pull/46530
 * https://github.com/tgstation/tgstation/pull/48831
 * https://github.com/tgstation/tgstation/pull/52415 **|| Code comments were added here**
.
 * https://github.com/tgstation/tgstation/pull/52417 **|| Verb to see all timers. It can help with solving some lag issues in case we will have any**
.
 * https://github.com/tgstation/tgstation/pull/53217
 * https://github.com/tgstation/tgstation/pull/54977
 * https://github.com/tgstation/tgstation/pull/55140
 * https://github.com/tgstation/tgstation/pull/55803
 * https://github.com/tgstation/tgstation/pull/57588
 * https://github.com/tgstation/tgstation/pull/59718
 * https://github.com/tgstation/tgstation/pull/59284


code/__defines/subsystems.dm
```diff
@@ -43,6 +47,15 @@
...
 }
 
+/**
+	Create a new timer and add it to the queue.
+	* Arguments:
+	* * callback the callback to call on timer finish
+	* * wait deciseconds to run the timer for
+	* * flags flags for this timer, see: code\__DEFINES\subsystems.dm
+*/
+#define addtimer(args...) _addtimer(args, file = __FILE__, line = __LINE__)
```
NanakoAC, line 54: this macro already existed, has it been moved or remade or duplicated?
does this version add anything meaningful?

code/__defines/time.dm
```diff
@@ -83,9 +83,13 @@ GLOBAL_LIST_INIT(month_seconds_cumulative_leap, list(
 ))
 
 
+#define TICKS *world.tick_lag
```
NanakoAC, line 86: these three defines need explanation, what are they ? what do they do?

code/controllers/subsystems/timer.dm
```diff
@@ -144,198 +225,203 @@ SUBSYSTEM_DEF(timer)
...
 
-	name = "Timer: " + num2text(id, 8) + ", TTR: [timeToRun], Flags: [jointext(bitfield2list(flags, list("TIMER_UNIQUE", "TIMER_OVERRIDE", "TIMER_CLIENT_TIME", "TIMER_STOPPABLE", "TIMER_NO_HASH_WAIT")), ", ")], callBack: \ref[callBack], callBack.object: [callBack.object]\ref[callBack.object]([getcallingtype()]), callBack.delegate:[callBack.delegate]([callBack.arguments ? callBack.arguments.Join(", ") : ""])"
+	if ((timeToRun < world.time || timeToRun < SStimer.head_offset) && !(flags & TIMER_CLIENT_TIME))
+		CRASH("Invalid timer state: Timer created that would require a backtrack to run (addtimer would never let this happen): [SStimer.get_timer_debug_string(src)]")
 
-	if (callBack.object != GLOBAL_PROC)
+	if (callBack.object != GLOBAL_PROC && !QDESTROYING(callBack.object))
 		LAZYADD(callBack.object.active_timers, src)
 
-	if (flags & TIMER_CLIENT_TIME)
```
NanakoAC, line 255: has this client time feature been removed? Why?

code/__defines/subsystems.dm
DTraitor: > Turns /proc/addtimer into /proc/_addtimer and creates an addtimer macro in its place. This is because the macro will pass in the file and line of what is creating the timer. I wanted to use gib_stack_trace() and friends, but those regularly turned up blank exception descriptions, and were useless. If there's anything better, let me know, but for now this is better than nothing.
>
> The command will only show those still in the buckets. This is to avoid both leaving too much of a footprint on the addtimer proc and due to memory concerns of logging all timers being created.

– PR where it was added. 
It's used to track where timer was added.

code/__defines/time.dm
DTraitor: It's doing the same thing as #define SECONDS. Just to make code more readable. I probably need to replace some *world.tick_lag with a define

code/controllers/subsystems/timer.dm
DTraitor: Whoops, It seems I accidentally deleted it because it exists in tg SStimer

code/datums/datum.dm
```diff
@@ -28,7 +28,7 @@
 	active_timers = null
 	for(var/thing in timers)
 		var/datum/timedevent/timer = thing
-		if (timer.spent)
+		if (timer.spent && !(timer.flags & TIMER_DELETE_ME))
```
NanakoAC: whats with this flag? are we going to need to add that in everywhere now to ensure timers get deleted properly? Why would that not be default behaviour

NanakoAC (commented):
some concerns and questions
the delete flag worries me most, existing content may need significant adapting for that, i question its utility

code/datums/datum.dm
DTraitor: You shouldn't worry about that, it's used to qdelete TIMER_LOOP

code/__defines/time.dm
DTraitor: TICKS doing the same thing as SECONDS;

DS2TICKS mean "delay seconds to ticks". It was originally used in /proc/stoplag() (used instead of sleep(1) or sleep() ) but we don't have it and I'm not going to port it (at least not in this PR);

TICKS2DS mean "ticks to delay seconds".

code/controllers/subsystems/timer.dm
DTraitor: It's still there, just looks a bit different.
`timeToRun = (flags & TIMER_CLIENT_TIME ? REALTIMEOFDAY : world.time) + wait`

code/datums/datum.dm
NanakoAC: i don't quite follow, nothing about loops is being checked here.

It seems to govern whether timers are deleted when the parent datum is. I cant think of any situations we wouldn't want to do that, 
how is it limited to loops?

DTraitor: removed that flag

NanakoAC (approved):

PR #1575, opened 2021-08-19
Line profiler

DTraitor:
### Purpose
Ports: 
https://github.com/vgstation-coders/vgstation13/pull/16529   // They changed few things so I used it.
https://github.com/tgstation/tgstation/pull/48672

### There is an example how to use it in vgstation PR. Make sure that you've added necessary defines to procs you are going profile.
It **MAY** conflict with SStimer PR because of SHORT_REAL_LIMIT define but it shouldn't since it has the same place.


NanakoAC (approved):
this should probably remain disabled in production code, its just a feature for local debugging

PR #1576, opened 2021-08-19
Small clears in main folders.

DeadLineSS13:
# About this PRs.
Sorting, removing things that we don't need.


.dockerignore
```diff
@@ -12,6 +12,5 @@ Dockerfile
 .editorconfig
 .gitattributes
 .gitignore
-.travis.yml
```
NanakoAC, line 15: don't we use travis? it runs checks

DeadLineSS13: We used CI system run checks. Travis old thing and don't supported yet.

NanakoAC (approved):

PR #1578, opened 2021-08-20
Update asset_cache system to last version.

DeadLineSS13:
# About this PR
All needed information in asset_cache folder Readme.md


DeadLineSS13:
Asset_cache updated TG system with assets for html, css and js. Basically, it is needed for the stable operation of TGUI, and it also solves the problem with NanoUI loading 99%. And asset_cache support downloading from the CDN cache. That is, we can reduce the unloading from the server using the CDN server. One of the possibilities.

DeadLineSS13:
And main feature that asset_cache will send only validated html, css and js in build. That is, we protect players from third-party hacking through the cache.

code/datums/craft/menu.dm
```diff
@@ -53,7 +53,7 @@
 	if(CR)
 		data["cur_item"] = list(
 			"name" = CR.name,
-			"icon" = getAtomCacheFilename(CR.result),
```
NanakoAC, line 56: This was doing more than just sanitize, this does not seem like an adequate replacement. whats the need for this change?

code/datums/craft/menu.dm
```diff
@@ -69,6 +69,8 @@
 
 	ui = SSnano.try_update_ui(user, src, ui_key, ui, data, force_open)
 	if (!ui)
+		var/datum/asset/assets = get_asset_datum(/datum/asset/simple/craft)
```
NanakoAC, line 72: this seems like way too much to do while loading a UI, why is this here? Resending the assets every time the menu is opened doesn't make sense

code/datums/craft/recipe.dm
```diff
@@ -66,12 +66,12 @@
 /datum/craft_recipe/proc/get_step_descriptions(var/skip = 0)
 	var/list/data = list()
 	for(var/datum/craft_step/CS in passive_steps)
-		data += list(list("icon" = getAtomCacheFilename(CS.icon_type), "desc" = CS.desc))
+		data += list(list("icon" = sanitizeFileName("[CS.icon_type].png"), "desc" = CS.desc))
```
NanakoAC, line 69: as previous comment about getAtomCacheFilename

code/game/machinery/autolathe.dm
```diff
@@ -182,6 +182,8 @@
 
 	ui = SSnano.try_update_ui(user, src, ui_key, ui, data)
 	if (!ui)
+		var/datum/asset/assets = get_asset_datum(/datum/asset/simple/research_designs)
```
NanakoAC, line 185: as previous concern about this

NanakoAC (commented):
some concerns
sending assets in ui opening seems redundant and a lot of extra load

code/datums/craft/menu.dm
DeadLineSS13: > This was doing more than just sanitize, this does not seem like an adequate replacement. whats the need for this change?

Because getAtomCacheFilename is meaningless code in this version of asset_cache.

code/datums/craft/menu.dm
DeadLineSS13: > this seems like way too much to do while loading a UI, why is this here? Resending the assets every time the menu is opened doesn't make sense

Because when creating a UI, assets are also sent, but for UI. It won't kill us. Since it is checked whether the player has such a file and hasch code.

NanakoAC: Checking all those files seems like a lot of wasted effort though. Afaik the ui is created whenever a player opens a menu

This seems like i's adding more performance cost to that, which was already a bit sluggish

DeadLineSS13: > Checking all those files seems like a lot of wasted effort though. Afaik the ui is created whenever a player opens a menu
> 
> This seems like i's adding more performance cost to that, which was already a bit sluggish

Every time a new window is created. An asset with the desired template is sent.
https://github.com/DS-13-Dev-Team/DS13/blob/653ccbff9cf52b836f863be1d835bdafdc77fd7e/code/modules/nano/nanoui.dm#L100

NanakoAC (approved):

PR #1583, opened 2021-08-21
Replace default Byond logs and revesion to RUST_G

DeadLineSS13:
## About this PR
Used a lof of tgstation/rust-g things.


DeadLineSS13:
RUST_G this thing already was added to the Repo. But basically this thing is needed for the stable operation of TGS. But the advantage is that it writes information to logs faster and issues a revision faster.

https://github.com/tgstation/rust-g

PR #1596, opened 2021-08-24
Fixed some timers, few things in research and Open RIG Interface button

DTraitor:
### Purpose
Fixed some timers not having TIMER_STOPPABLE flag. Circuit imprinter recognizes reagents again. Seeker and mining tech no longer result in runtimes. Replaced oxygen with water in incend ammo design because it appears that **OXY**phoron requires water to make. Javelin racks can be printed. Open RIG Interface verb now also works if you have 2 or more RIGs in hands and also appears in dropdown list (rmb click).


DTraitor:
I'd like this to be reviewed and merged before we will launch server today. Yesterday people complained about RnD a bit

NanakoAC (approved):

PR #1620, opened 2021-09-04
Infector

NanakoAC:
changes: 
  - experiment: "Added the Infector, a complex new support necromorph with abilities to convert humans, construct corruption nodes, heal and buff other necromorphs."
  - rscadd: "Added necrotoxin, a poison that converts its victims into necromorphs if it kills them. Necrotoxin is applied by all infector attacks."
  - tweak: "Reworked chemical metabolism rates. The base rate is a bit slower, but now chemicals metabolize faster, the larger the quantity is in a patient's system."


Ketrai:
OKAY AFTER TESTING.
- [x] Spine launch uses the description for long shot cooldown.
- [x] You can target yourself and shoot yourself.
- [x] Mend gives a placeholder description when selected.
- [x] Flap can throw humans into walls for massive damage.

- [x]  Execution bugs. 
- [x] Infector sprite stays small.
- [ ] when grappled and infection finishes, it will bug out massively. Causing a human/necromorph that has the necro's abilities it was supposed to turn into. But keeps a human sprite. Also blind/immobile.
- [ ] Due to coding, infected humans can't see messages in necrochat.
- [x] Sprite can get offset, reset the offset properly.
- [x] Infected humans keep their armor overlays. Which looks odd.

- [ ] Necrotoxin can never trigger anyone to turn. Maybe make it check for toxins > 50- [ ]100? Because you'll die of brain/oxy damage 100% of the time.
- [x] Necrotoxin has no buildup. 5 units is just as effective as 500.
- [x] Flap and parasitic leap share a cooldown variable/define. Using either puts them on the same cooldown.
- [x] **NECROS CAN BE TARGETED BY PARASITIC LEAP**
- [x] Reanimate uses the same description as engorge.
- [x] If the target moves whilst being engorged, the ability cancels and essence is lost.

Other notes and suggestions.
- [x] Essence income is really slow. Consider using a higher max cap, like 150. But making essence income 10x as fast. In return, all abilities roughly cost 10 times as much as well. But now it's easier to read and estimate. Nodes cost an insane amount of time before you can do anything. And small structures are too expensive compared to just spamming nodes. It takes a signal 150 seconds to make a root/branch. It takes the infestor 240 seconds. Which is far longer, even if the node is stronger. Growth note for 30 essence is pretty infeasible. 2 infectors would needs 15 minutes to put one out. At that point it's faster to just spam propagators unless you want corruption in the odd butt- []end of the ship. Maybe replace it by a node that needs 5 minutes to activate, and about 5- []10 minutes to be earned. Don't forget, expensive structures take long to build too, which just adds to the time.

NanakoAC:
RE: Essence income, i've added a change where the income is accelerated while you're above 50%, to encourage saving and cautious spending

NanakoAC:
>  Due to coding, infected humans can't see messages in necrochat.
> Necrotoxin can never trigger anyone to turn. Maybe make it check for toxins > 50- [ ]100? Because you'll die of brain/oxy damage 100% of the time.


Unable to reproduce these items

Issue #1628, opened 2021-09-07
More infector bugs.


Ketrai:
- [x] Infectors can still use parasitic leap on other necromorphs. (when they're standing or buckled
- [x] spine shot has no cooldown
- [x] DOUBLE CHECK TO MAKE SURE THEY DON'T SPAWN EXTRA BIOMASS WHEN THEIR NECROS DIE. (Although this might be because of execute, and the fact conversions spawn gibs. Consider lowering the execution bonus to 2KG or something, the gibs are usually 4-8 KG already anyways.
- [x] They can drag during a flap, which is obviously unfun for humans and necros alike
- [x] decapitated humans can be turned via reanimate
- [x] engorge can be cast on already engorged necromorph. Does nothing, retracts bio
- [ ] reanimate can be cancelled, which is fine, but it doesn't refund essence.
- [ ] Can't heal missing limbs with mend
- [x] mend still uses the engorge description... did you overwrite your own changes?

SUGGESTION.

- [x] Mend is more expensive than reanimate on average. Would it be possible to cure 1/3rd the amount healed as a reduction in lasting damage taken to make it more worthwhile? Currently only the marker can do it, and wasting a rebuild on a robust infector or enhanced is quite wasteful.

NanakoAC:
>reanimate can be cancelled, which is fine, but it doesn't refund essence.
Cannot reproduce this, refund is working fine on cancellation

PR #1632, opened 2021-09-08
Infector Fixes I

NanakoAC:
changes: 
  - experiment: "All mobs now take 50% more damage from all sources while they are performing an execution. Its a vulnerable time for others to get you while your back is turned!"
  - tweak: "Reduced the infector's movespeed, max health, and melee damage+armorpen."
  - tweak: "Infector can no longer grapple or pull humans. It has no limbs to do so!"
  - tweak: "Mend can now heal Lasting Damage."
  - bugfix: "Fixed infector spine shot having no cooldown."
  - bugfix: "Parasite Leap can no longer hit other necromorphs."
  - bugfix: "Fixed being able to reanimate headless corpses."
  - bugfix: "Engorge can no longer be used again on already enlarged targets."

### DIFF
--- a/code/modules/mob/living/abilities/execution/execution.dm
+++ b/code/modules/mob/living/abilities/execution/execution.dm
@@ -37,6 +37,8 @@ if (result == EXECUTION_CANCEL && can_interrupt){\
 	interrupt();\
 	return}
 
+#define EXECUTION_DAMAGE_VULNERABILITY	1.5
+
 //Execution status vars
 #define STATUS_NOT_STARTED	0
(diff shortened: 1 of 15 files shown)


NanakoAC:
Fixes most of the bugs in #1628

reklaw9:
If i remember correctly, they can infect headless bodies, but their "transformation" is slower than normal.

The whole reason that infectors target the brain when injecting their bacteria is to kill the brain very fast so the body's immune system stops working so the "necro bacteria" can fully infect it(it also apparently has something to do with the spine, but idk).

PR #1633, opened 2021-09-08
SSoverlays, attemp 3.5

DTraitor:
All the info: #1570


NanakoAC:
why is this 12 files when the last one was almost 300 ?

DTraitor:
Because it's just a subsystem this time. Going to replace 'overlays |=' with procs later

NanakoAC (approved):

PR #1642, opened 2021-09-09
Fixed layers decal and disposable pipe

DeadLineSS13:
## About this PR
Changed layers to needed var.

### DIFF
--- a/code/modules/recycling/disposal.dm
+++ b/code/modules/recycling/disposal.dm
@@ -832,7 +832,7 @@
 
 	src.set_invisibility(101)	// make invisible (since we won't delete the pipe immediately)
 	var/obj/structure/disposalholder/H = locate() in src
-	if(H)
+	if(!QDELETED(H))
 		// holder was present
 		H.active = 0
 		var/turf/T = src.loc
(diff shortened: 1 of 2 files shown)


code/modules/recycling/disposal.dm
```diff
@@ -675,258 +675,259 @@
...
+/obj/structure/disposalpipe/proc/broken(var/remains = 0)
+	if(remains)
+		for(var/D in GLOB.cardinal)
+			if(D & dpdir)
+				var/obj/structure/disposalpipe/broken/P = new(src.loc)
+				P.set_dir(D)
+
+	src.set_invisibility(101)	// make invisible (since we won't delete the pipe immediately)
+	var/obj/structure/disposalholder/H = locate() in src
+	if(H)
```
NanakoAC: There are a LOT of places here where a disposalholder can get deleted

this should check if (!QDELETED(H)) to ensure its not getting deleted multiple times

code/modules/recycling/disposal.dm
```diff
@@ -675,258 +675,259 @@
...
-					AM.forceMove(T)
-					AM.pipe_eject(0)
-				qdel(H)
-				return
+			expel(H, T, 0)
 
-			// otherwise, do normal expel from turf
-			if(H)
-				expel(H, T, 0)
+	spawn(2)	// delete pipe after 2 ticks to ensure expel proc finished
```
NanakoAC: use QDEL_IN instead of spawn

NanakoAC (changes requested):
a few tiny tweaks

NanakoAC (approved):

PR #1648, opened 2021-09-10
Removed some useless if (is_processing) checks and some debug messages

DTraitor:

### DIFF
--- a/code/datums/extensions/movement/rotate_facing.dm
+++ b/code/datums/extensions/movement/rotate_facing.dm
@@ -48,8 +48,7 @@
 	target = newtarget
 	GLOB.moved_event.register(target, src, target_moved())
 
-	complete_cycles = 0
-	START_PROCESSING(SSfastprocess, src)
+	start_tracking()
 
 
 /*
@@ -96,18 +95,31 @@
 	.= AM.get_turf_at_pixel_offset(rotated_forward_vector)
 	release_vector(rotated_forward_vector)
 
-/datum/extension/rotate_facing/proc/target_moved()
+/*
+	This is called when a new target is set
+	In active track mode, it is also called when the holder or target moves
+	This starts the processing
+*/
+/datum/extension/rotate_facing/proc/start_tracking()
 	complete_cycles = 0
 	START_PROCESSING(SSfastprocess, src)
 
+/*
+	This is called once we are facing the target, we'll stop processing until more turning is needed
+*/
+/datum/extension/rotate_facing/proc/end_tracking()
+	STOP_PROCESSING(SSfastprocess, src)
+
+/datum/extension/rotate_facing/proc/target_moved()
+	start_tracking()
+
 /datum/extension/rotate_facing/proc/holder_moved()
-	complete_cycles = 0
-	START_PROCESSING(SSfastprocess, src)
+	start_tracking()
 
 
 /datum/extension/rotate_facing/Process()
 	if (QDELETED(target))
-		STOP_PROCESSING(SSfastprocess, src)
+		end_tracking()
 		return
 
 	//Where are we currently looking?
@@ -156,7 +168,7 @@
 		complete_cycles++
 		//If we've been at the target for two ticks in a row, we're done
 		if (complete_cycles > 1)
-			STOP_PROCESSING(SSfastprocess, src)
+			end_tracking()
 	else
 		complete_cycles = 0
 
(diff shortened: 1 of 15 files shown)


code/datums/extensions/movement/rotate_facing.dm
```diff
@@ -47,7 +47,9 @@
 
 	target = newtarget
 	GLOB.moved_event.register(target, src, target_moved())
-	start_tracking()
```
NanakoAC, line 50: this is a step backwards, please revert changes to this file.

Wrapper functions like this are important for future extensibility, and they're useful right now if there's more than one line of code

code/game/gamemodes/crew_objective/crew_objective.dm
```diff
@@ -362,8 +362,7 @@
 */
 /datum/crew_objective/proc/fade()
 	status	=	CO_STATUS_FADING
-	if (is_processing)
```
NanakoAC, line 365: why remove this check?

code/game/gamemodes/meteor/meteor_object.dm
```diff
@@ -66,7 +66,7 @@ GLOBAL_LIST_EMPTY(asteroids)
 /proc/register_asteroid(var/obj/effect/meteor/M)
 	GLOB.asteroids += M
 	if (GLOB.asteroidcannon)
-		GLOB.asteroidcannon.fire_handler.wake_up()
```
NanakoAC: don't remove wrapper functions

code/game/objects/objs.dm
```diff
@@ -31,8 +31,7 @@
 /obj/Destroy()
 	if (is_processing)
 		var/datum/controller/subsystem/processing/P = processing_subsystems_by_varname[is_processing]
-		if (P)
```
NanakoAC, line 34: what is going to happen if the subsystem isn't there without this check?

code/modules/clothing/spacesuits/rig/modules/ds13/kinesis/kinesis.dm
```diff
@@ -505,18 +505,6 @@
 
 	The kinesis module uses fastprocess, ticking 5 times per second
 */
-/obj/item/rig_module/kinesis/proc/start_processing()
```
NanakoAC, line 508: wrapper functions with process in the name should be renamed, but still kept

code/modules/projectiles/firemodes/firemode_remote.dm
```diff
@@ -67,8 +67,7 @@
 
 		update_tethers()
 
-	if (!is_processing)
-		START_PROCESSING(SSfastprocess, src)
```
NanakoAC, line 71: what happens if you try to start processing multiple times?

NanakoAC (changes requested):
A lot of these changes don't make sense, some require farther explanation, some need modified or reverted

code/game/gamemodes/crew_objective/crew_objective.dm
DTraitor: Because it checks if it is processing in STOP_PROCESSING()

code/game/objects/objs.dm
DTraitor: If subsystem isn't there then we are doing something wrong and this should be fixed. Probably it'll result in a runtime

code/modules/projectiles/firemodes/firemode_remote.dm
DTraitor: it'll result in a runtime.

code/game/objects/objs.dm
DTraitor: I think that P is useless. is_processing already contains enough info to STOP_PROCESSING

code/modules/clothing/spacesuits/rig/modules/ds13/kinesis/kinesis.dm
DTraitor: Using proc instead of calling START_PROCESSING is useless here. It's called only in one place. Same applies to stop_processing() proc here

DTraitor:
STOP_PROCESSING(is_processing, src) now works
image

code/game/machinery/asteroid_cannon/ads_extension.dm
```diff
@@ -117,4 +117,11 @@
...
 		return FALSE
 	gun = holder
-	START_PROCESSING(SSfastprocess, src)
+	wake_up()
+
+
+
+
+/datum/extension/asteroidcannon/proc/wake_up()
+	if (!is_processing)
```
NanakoAC: as you demonstrated, this check is unnecessary since it already does that check as part of START_PROCESSING and STOP_PROCESSING

so you can remove that and others like it

NanakoAC (approved):

PR #1663, opened 2021-09-14
Move all screen objects type humans to hud.

DeadLineSS13:
## About thi PR.
This first step rework Screen. We moved these, beacause we don't need mixed var hud with var mob.

### DIFF
--- a/code/_helpers/mobs.dm
+++ b/code/_helpers/mobs.dm
@@ -305,7 +305,7 @@ var/datum/callback/proc_to_call, var/proc_interval = 10)
 				//We will pass: User, Interval
 				proc_to_call.Invoke(user, target, proc_interval)
 
-		if(target_zone && user.hud_used.zone_sel.selecting != target_zone)
+		if(target_zone && get_zone_sel(user) != target_zone)
 			. = 0
 			break
 
--- a/code/_helpers/mobs.dm
+++ b/code/_helpers/mobs.dm
@@ -226,13 +226,13 @@ var/datum/callback/proc_to_call, var/proc_interval = 10)
 
 /proc/doafter_blocked(user)
 	.=FALSE
-	if (has_extension(user, /datum/extension/interrupt_doafter))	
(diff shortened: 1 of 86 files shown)


code/_helpers/mobs.dm
```diff
@@ -305,7 +305,7 @@ var/datum/callback/proc_to_call, var/proc_interval = 10)
 				//We will pass: User, Interval
 				proc_to_call.Invoke(user, target, proc_interval)
 
-		if(target_zone && user.zone_sel.selecting != target_zone)
+		if(target_zone && user.hud_used.zone_sel.selecting != target_zone)
```
NanakoAC: Replace this part with get_zone_sel(user), its a proc to minimise this reference walking

code/_helpers/unsorted.dm
```diff
@@ -963,7 +963,7 @@ var/global/list/common_tools = list(
 		. = TRUE
 
 	if(M == user)
-		var/hitzone = check_zone(user.zone_sel.selecting)
+		var/hitzone = check_zone(user.hud_used.zone_sel.selecting)
```
NanakoAC: get_zone_sel here too

code/game/gamemodes/changeling/changeling_powers.dm
```diff
@@ -710,7 +710,7 @@ var/list/datum/absorbed_dna/hivemind_bank = list()
 	if(!(T in view(changeling.sting_range))) return
 	if(!sting_can_reach(T, changeling.sting_range)) return
 	if(!changeling_power(required_chems)) return
-	var/obj/item/organ/external/target_limb = T.get_organ(src.zone_sel.selecting)
+	var/obj/item/organ/external/target_limb = T.get_organ(src.hud_used.zone_sel.selecting)
```
NanakoAC, line 713: get_zone_sel here too, and all other uses

NanakoAC (changes requested):

NanakoAC (approved):

PR #1669, opened 2021-09-15
Fixes DreamLanguage

DeadLineSS13:
## About this PR
Fixed most problems founded by DM-Language. Now we have 116 problems.

## Changelog
🆑 DeadLineSS13
fix: Fixed layers UI invetory with dark layers. 
/🆑


code/_helpers/global_access.dm
```diff
@@ -19,86 +19,10 @@
 			return global.DEPT_FREQS;
 		if("Debug2")
 			return global.Debug2;
-		if("Failsafe")
-			return global.Failsafe;
-		if("GLOB")
```
NanakoAC, line 24: isn't this constantly used everywhere?  so are the subsystems

code/game/gamemodes/crew_objective/objective_event.dm
```diff
@@ -27,7 +27,7 @@
 */
 /datum/event/crew_objective
 	startWhen = 0
-	endWhen = NEAR_INFINITY	//This won't end naturally, only when crew finish the objective,
```
NanakoAC, line 30: why was the near infinity removed?

code/modules/clothing/spacesuits/rig/modules/ds13/kinesis/extension.dm
```diff
@@ -13,8 +13,8 @@
 	var/cached_plane
 
 	//The graphical filter we assigned to the held item
-	var/dm_filter/ripple
```
NanakoAC: no don't do this, dm_filter is a valid native type, just hidden
there should be a file somewhere which declares it to prevent compiler errors, i don't know why dreamchecker isn't seeing that

code/modules/clothing/spacesuits/rig/modules/ds13/kinesis/extension.dm
```diff
@@ -32,7 +32,7 @@
 	ripple = subject.filters[subject.filters.len]
 	animate(ripple, radius = 16, size = 1, time = 4, loop = -1, flags = ANIMATION_PARALLEL)
 
-	outline = filter(type = "outline", size = 3, color = COLOR_KINESIS_INDIGO, alpha = 128)
```
NanakoAC, line 35: this is a functional change, revert

code/modules/mob/living/abilities/charge.dm
```diff
@@ -82,7 +82,7 @@
 	var/cached_step_interval
 
 	var/blur_filter_strength = 2
-	var/dm_filter/blur
```
NanakoAC: likewise, dm_filter is valid, so for all similar changes

NanakoAC (changes requested):
a few concerns with dm filter and global vars

code/_helpers/global_access.dm
DeadLineSS13: We don't need because GLOBAL system already do it thing.

code/game/gamemodes/crew_objective/objective_event.dm
DeadLineSS13: NEAR_INFINITY <=> INFINITY

code/modules/clothing/spacesuits/rig/modules/ds13/kinesis/extension.dm
DeadLineSS13: filter type outline don't support alpha. Check DM references filter outline

DTraitor: It does, I've tested

DeadLineSS13:
## Changelog

🆑 DeadLineSS13
fix: Fixed layers UI invetory with dark layers. 
/🆑

NanakoAC (approved):

PR #1681, opened 2021-09-15
Encumbrance Refactor

NanakoAC:
This PR redesigns how equipment encumbrance works so that it is only recalculated when equipped things change, instead of being calculated EVERY SINGLE TIME A HUMAN TAKES A STEP

this should yield a minor performance improvement, and is another step of many in improving movement
This PR cuts out about 10% of the wasted work in movement_delay, still lots more to do

In addition, encumbrance now affects attackspeed and evasion, and is now decreased by your hauling skill. Put points in that skill to reduce the penalties from heavy armor


Snypehunter007:
Change log for the last thing?

PR #1709, opened 2021-09-19
Fixes huds.

DeadLineSS13:
## About the PR
Fixed 2 runtimes with hud_used.

## Changelog
🆑 DeadLineSS13
fix: Fixed 2 runtimes with hud_used.
/🆑

### DIFF
--- a/code/_onclick/hud/meter/health.dm
+++ b/code/_onclick/hud/meter/health.dm
@@ -28,8 +28,7 @@
 /obj/screen/meter/health/set_mob(var/mob/living/newmob)
 	.=..()
 	GLOB.updatehealth_event.register(L, src, /obj/screen/meter/proc/update)
-	if(L?.hud_used)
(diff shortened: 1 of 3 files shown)


code/_onclick/hud/meter/health.dm
```diff
@@ -28,7 +28,8 @@
 /obj/screen/meter/health/set_mob(var/mob/living/newmob)
 	.=..()
 	GLOB.updatehealth_event.register(L, src, /obj/screen/meter/proc/update)
-	L.hud_used.hud_healthbar = src
```
NanakoAC: this isn't a fix

The healthbar is part of the hud now, you've got to either make sure this called after the hud is set up. That var needs to be populated

NanakoAC (changes requested):

NanakoAC (approved):

PR #1713, opened 2021-09-20
Adds in Max Stone's custom advanced RIG suit

KeRSedChaplain:
As it says on the tin.

Unsure why there's commits from DTraitor on this but it's just a line deletion so it should be harmless


NanakoAC (approved):
this just adds it to assets, its not implemented or obtainable here by the look of it.

Snypehunter007:
> 
> 
> this just adds it to assets, its not implemented or obtainable here by the look of it.

No, it is definitely implemented.

PR #1726, opened 2021-09-23
Small screenrework

DeadLineSS13:
## About this PR
Reworked screen and hud.

## Changelog
🆑	DeadLineSS13
refactor: Reworked screen works.
refactor: Reworked hud works.
/🆑


code/__defines/misc.dm
```diff
@@ -1,4 +1,8 @@
 
+//Fullscreen overlay resolution in tiles.
+#define FULLSCREEN_OVERLAY_RESOLUTION_X 15
```
NanakoAC: this seems bad, these should be calculated from client's viewsize not hardcoded

code/_helpers/client.dm
```diff
@@ -31,47 +31,49 @@
...
-				var/mob/living/L = mob
-				L.handle_regular_hud_updates(FALSE)//Pass false here to not call update vision and avoid an infinite loop
-
-				//Update hud healthbar if one exists, so that its clamped to screen size
-				if (L.hud_used.hud_healthbar)
-					L.hud_used.hud_healthbar.set_size(TRUE)
-
-		spawn()
-			view = temp_view
+/client/proc/change_view(new_size)
```
NanakoAC, line 34: this proc name is less descriptive, keep the old name

and why fewer vars? You've removed some functionality here
It took a LOT of work to get this thing working correctly, i'm suspicious of changes

code/_helpers/client.dm
```diff
@@ -31,47 +31,49 @@
...
+		mob.client.update_skybox(TRUE)
+	if (isliving(mob))
+		var/mob/living/L = mob
+		L.handle_regular_hud_updates(FALSE)//Pass false here to not call update vision and avoid an infinite loop
+
+		//Update hud healthbar if one exists, so that its clamped to screen size
+		if (L.hud_used.hud_healthbar)
+			L.hud_used.hud_healthbar.set_size(TRUE)
+
+	if (prefs.auto_fit_viewport)
```
NanakoAC, line 67: what is this?

code/_helpers/client.dm
```diff
@@ -31,47 +31,49 @@
...
+		addtimer(CALLBACK(src,.verb/fit_viewport,10)) //Delayed to avoid wingets from Login calls.
 
 //Returns the width of the viewport/map window, in pixels
 /client/proc/get_viewport_width()
-	return ((temp_view*2)+1) * WORLD_ICON_SIZE
+	return ((view_radius*2)+1) * WORLD_ICON_SIZE
```
NanakoAC, line 72: this temp view var was necessary, do you understand what it was used for before you remove it? How have you resolved its problems?

code/_onclick/click_catcher.dm
```diff
@@ -1,127 +1,62 @@
-/*
-	A click catcher is an invisible object on the client screen which draws below pretty much everything else, including turfs.
```
NanakoAC, line 2: why have you removed all this documentation and helper procs?

code/_onclick/click_catcher.dm
```diff
@@ -1,127 +1,62 @@
...
 //Mob Procs
 //-------------
 /mob/proc/add_click_catcher()
-	if (client)
-		client.remake_click_catcher()
-
-/mob/proc/remove_click_catcher()
-	if (client)
-		client.remove_click_catcher()
+	client.screen += client.void
```
NanakoAC, line 5: void is not a descriptive name, call it click_catcher

Remaking it here was necessary to avoid view size problems, how have you resolved that?

code/_onclick/click_catcher.dm
```diff
@@ -1,127 +1,62 @@
...
+	icon = 'icons/hud/screen_gen.dmi'
+	icon_state = "catcher"
 	plane = CLICKCATCHER_PLANE
-	mouse_opacity = 2
-	screen_loc = "CENTER-7,CENTER-7"
+	mouse_opacity = MOUSE_OPACITY_OPAQUE
+	screen_loc = "CENTER"
+
+
+#define MAX_SAFE_BYOND_ICON_SCALE_TILES (MAX_SAFE_BYOND_ICON_SCALE_PX / world.icon_size)
```
NanakoAC, line 21: these require explanation, maximum safe?

code/_onclick/click_catcher.dm
```diff
@@ -1,127 +1,62 @@
...
+#define MAX_SAFE_BYOND_ICON_SCALE_PX (33 * 32)			//Not using world.icon_size on purpose.
+
+
+/obj/screen/click_catcher/proc/UpdateGreed(view_size_x = 15, view_size_y = 15)
```
NanakoAC, line 25: greed? this needs a better name and explanation

code/_onclick/hud/fullscreen.dm
```diff
@@ -1,125 +1,104 @@
...
 
 /mob/proc/hide_fullscreens()
 	if(client)
 		for(var/category in screens)
 			client.screen -= screens[category]
 
-/mob/proc/reload_fullscreen()
+/mob/proc/reload_fullscreens()
 	if(client)
+		var/obj/screen/fullscreen/screen
```
NanakoAC, line 62: why are you doing this here? it should be inside the loop

code/_onclick/hud/fullscreen.dm
```diff
@@ -1,125 +1,104 @@
...
-	var/mob/owner
-
-/obj/screen/fullscreen/proc/set_size(var/client/C)
-	//Here we select (and if needed, generate) the icon for the right size
-	if (C.temp_view == world.view)
-		return	//No special sizing needed
-
-	icon = get_or_create_fullscreen(C.temp_view)
-
+	var/fs_view = WORLD_VIEW
```
NanakoAC, line 78: no abbreviations please, give this variable a full title descriptive of what it does

code/_onclick/hud/hud.dm
```diff
@@ -23,6 +24,15 @@
 	var/obj/screen/move_intent
 	var/obj/screen/stamina/stamina_bar
 	var/obj/screen/meter/health/hud_healthbar
+	var/obj/screen/meter_component/current/remaining_meter	//The actual remaining health, in red or green
+	var/obj/screen/meter_component/delta/delta_meter	//A yellow section indicating recent loss
+	var/obj/screen/meter_component/limit/limit_meter	//A solid grey block at the end, representing reduced maximum
+	var/obj/screen/meter_component/text/textholder
+	var/obj/screen/meter/resource/hud_resource
```
NanakoAC, line 27: this is really bad, we can't be having all these variables
these meters need to be in a list, and the parent object responsible for all the components

i'm aware it was already kinda bad, but this makes it worse

code/_onclick/hud/human.dm
```diff
@@ -42,46 +41,93 @@
...
 	if(hud_data.has_healthbar)
 
-		using = new /obj/screen/meter/health(target.client)
-		src.adding += using
-		hud_elements |= using
+		hud_healthbar = new /obj/screen/meter/health(owner)
+		hud_healthbar.L = owner
+		infodisplay += hud_healthbar
+
+		remaining_meter = new /obj/screen/meter_component/current(hud_healthbar)
```
NanakoAC: this stuff definitely doesn't belong here, the meter creates its own components

code/_onclick/hud/meter/meter.dm
```diff
@@ -51,74 +50,26 @@
...
-/obj/screen/meter/proc/cache_data(var/atom/holder)
+/obj/screen/meter/proc/cache_data(atom/holder)
 
 
 /obj/screen/meter/Destroy()
 	QDEL_NULL(remaining_meter)
 	QDEL_NULL(delta_meter)
 	QDEL_NULL(limit_meter)
 	QDEL_NULL(textholder)
-	if (C)
```
NanakoAC, line 69: cleaning itself up from the client screen makes sense

code/_onclick/hud/meter/meter.dm
```diff
@@ -51,74 +50,26 @@
...
-		C.screen -= src
-		C = null
 	.=..()
 
-/obj/screen/meter/added_to_screen(var/client/newclient)
```
NanakoAC, line 74: this proc was definitely needed

code/_onclick/hud/meter/meter.dm
```diff
@@ -51,74 +50,26 @@
...
-	recreate_components(newclient)
-
-
-
-
-/obj/screen/meter/proc/recreate_components(var/client/newclient)
```
NanakoAC, line 80: so was this, i think most of the changes to meters will need reverted

code/_onclick/hud/screen_objects.dm
```diff
@@ -84,18 +147,86 @@
...
 	name = "intent"
-	icon = 'icons/mob/screen1_White.dmi'
+	icon = 'icons/hud/White.dmi'
 	icon_state = "intent_help"
 	screen_loc = ui_acti
 	var/intent = I_HELP
 
-/obj/screen/intent/Click(var/location, var/control, var/params)
+/obj/screen/intent/New(mob/C)
+	if (C)
```
NanakoAC, line 225: doing this in added_to_screen worked fine, putting code in new isnt a great idea. Especially when you didn't call parent!

code/game/atoms.dm
```diff
@@ -33,6 +33,8 @@
 	var/default_alpha = 255
 	var/default_scale = 1
 
+	var/list/filter_data //For handling persistent filters
```
NanakoAC: We were just arguing about memory footprint this morning. Adding a new list to all atoms is a terrible idea.

Extensions can manage filters for the few things that use them, no new list on atoms

code/game/objects/_items.dm
```diff
@@ -98,6 +98,12 @@
 	//Items are not dense typically
 	can_block_movement = FALSE
 
+	//Tooltip vars
+	///string form of an item's force. Edit this var only to set a custom force string
+	var/force_string
```
NanakoAC, line 103: string form of force? You mean the number that indicates how much damage it does? Why would you cache that

code/game/objects/_items.dm
```diff
@@ -98,6 +98,12 @@
...
+	var/last_force_string_check = 0
```
NanakoAC, line 104: use an extension for managing cooldowns and last checks if such is necessary

code/game/objects/_items.dm
```diff
@@ -98,6 +98,12 @@
...
+	var/tip_timer
```
NanakoAC, line 105: this needs explanation

code/game/objects/_items.dm
```diff
@@ -1005,4 +1007,72 @@ THIS SCOPE CODE IS DEPRECATED, USE AIM MODES INSTEAD.
 
 
 /obj/item/repair_needed()
-	return max_health - health
\ No newline at end of file
+	return max_health - health
+
+/obj/item/proc/set_force_string()
```
NanakoAC: this definitely doesn't need to be cached, just run the proc to fetch it on demand

and don't hardcode numbers here, use the same defines that weapon forces are derived from

code/game/objects/_items.dm
```diff
@@ -1005,4 +1007,72 @@ THIS SCOPE CODE IS DEPRECATED, USE AIM MODES INSTEAD.
...
+/obj/item/MouseDrop(atom/over, src_location, over_location, src_control, over_control, params)
+	. = ..()
+	remove_filter("hover_outline") //get rid of the hover effect in case the mouse exit isn't called if someone drags and drops an item and somthing goes wrong
+
+/obj/item/MouseExited()
+	deltimer(tip_timer) //delete any in-progress timer if the mouse is moved off the item before it finishes
+	closeToolTip(usr)
+	remove_filter("hover_outline")
+
+/obj/item/proc/apply_outline(outline_color = null)
```
NanakoAC: this puts an outline on anything someone hovers over? That seems like it would have a huge performance cost

is it restricted to things in range?

code/game/objects/_items.dm
```diff
@@ -1005,4 +1007,72 @@ THIS SCOPE CODE IS DEPRECATED, USE AIM MODES INSTEAD.
...
+	if(get(src, /mob) != usr || QDELETED(src) || isobserver(usr)) //cancel if the item isn't in an inventory, is being deleted, or if the person hovering is a ghost (so that people spectating you don't randomly make your items glow)
```
NanakoAC: this would be better accessed with a click handler that is only applied to conscious humans so that ghosts don't get this far in the first place

code/modules/client/client_defines.dm
```diff
@@ -1,10 +1,10 @@
 /client
-		//////////////////////
-		//BLACK MAGIC THINGS//
-		//////////////////////
-	parent_type = /datum
+	parent_type = /datum		// black magic
+	preload_rsc = PRELOAD_RSC	// This is 0 so we can set it to an URL once the player logs in and have them download the resources from a different server.
+	view = WORLD_VIEW
+	var/view_radius = WORLD_VIEW_NUM
```
NanakoAC: how does this differ from view? Doesn't it do the same thing temp_view did?

code/modules/client/client_defines.dm
```diff
@@ -1,10 +1,10 @@
...
+	var/datum/tooltip/tooltips
```
NanakoAC, line 6: why are tooltips cached in a list? Shouldn't they just be part of screen generally?

code/modules/client/client_procs.dm
```diff
@@ -110,6 +110,18 @@
 
 	..()	//redirect to hsrc.Topic()
 
+/client/proc/generate_clickcatcher()
```
NanakoAC, line 113: i dont see a need to seperate generate and apply here, we wouldn't want a click catcher that isn't sized properly, right?

code/modules/client/client_procs.dm
```diff
@@ -244,6 +263,11 @@
 	//DISCONNECT//
 	//////////////
 /client/Del()
+	if(!gc_destroyed)
```
NanakoAC: if (!QDELETED(src))

code/modules/client/client_procs.dm
```diff
@@ -252,9 +276,10 @@
 		GLOB.admins -= src
 	GLOB.ckey_directory -= ckey
 	GLOB.clients -= src
-	return ..()
 
-/client/Destroy()
+	QDEL_NULL(void)
+	QDEL_NULL(tooltips)
```
NanakoAC, line 281: QDEL_NULL_LIST

code/modules/client/preferences.dm
```diff
@@ -19,6 +19,14 @@ datum/preferences
 	var/ui_style = "Midnight"
 	var/tgui_fancy = TRUE
 	var/tgui_lock = FALSE
+	var/enable_tips = TRUE
+	var/tip_delay = 500 //tip delay in milliseconds
+
+	///Do we show item hover outlines?
+	var/itemoutline_pref = TRUE
```
NanakoAC, line 26: if this feature is going to be added it should not be optional, people would turn it off to get a gameplay advantage so nobody can see where they're looking

code/modules/client/verbs/ooc.dm
```diff
@@ -1,5 +1,67 @@
...
+	desired_width = min(desired_width, split_width - 300)
+
+	// Calculate and apply a best estimate
+	// +4 pixels are for the width of the splitter's handle
+	var/pct = 100 * (desired_width + 4) / split_width
+	winset(src, "mainwindow.split", "splitter=[pct]")
+
+	// Apply an ever-lowering offset until we finish or fail
+	var/delta
+	for(var/safety in 1 to 10)
```
NanakoAC, line 51: taking guesses in a loop seems hella sloppy, why is it doing this? Surely the right values can be calculated?

code/modules/mob/inventory.dm
```diff
@@ -379,4 +379,6 @@ var/list/slot_equipment_priority = list( \
 //placeholder until tg inventory system
 /mob/proc/is_holding(obj/item/I)
 	return ((I == l_hand) || (I == r_hand))
-z
\ No newline at end of file
+
+/mob/proc/can_equip(obj/item/I, slot, disable_warning = FALSE, bypass_equip_delay_self = FALSE)
```
NanakoAC: what's this new parameter used for? i don't think it belongs here

code/modules/mob/living/carbon/human/species/species_hud.dm
```diff
@@ -14,6 +14,7 @@
 	var/has_guns	=	TRUE	//Set to draw the gun/aiming toggle buttons
 	var/list/equip_slots = list() // Checked by mob_can_equip().
 	var/has_healthbar = FALSE	//Big healthbar across top of screen
+	var/has_resources = FALSE
```
NanakoAC, line 17: this var shouldn't exist, things might have more than one resource meter in future, they're designed to support an arbitrary quantity

code/modules/mob/mob_defines.dm
```diff
@@ -158,7 +158,7 @@
 
 	//Vision
 	var/view_offset = 0
-	var/view_range = 7
+	var/view_range = WORLD_VIEW_NUM
```
NanakoAC: rename this define WORLD_VIEW_RANGE

NanakoAC (changes requested):
This needs a lot of work.

This PR can't be merged as a whole, there are too many problems that each require significant work. It needs to be broken down into several smaller PRs

Everything you've done to meters is on the wrong track, each meter needs to be relatively self contained and manage its own thing, and in future things should be able to have arbitrary numbers of them, don't hardcode it into the design of hud code

The changes with filters, tooltips and force descriptions sound like nice features, but their implementation takes us down a path of unnecessary increase in memory footprint, i believe they can be done better.  Make use of extensions for things that will only apply occasionally, and click handlers to optimise the performance of it and limit who is dealing with those things

There are a lot of places where things need more descriptive names, comments, and general documentation

I'd like to see this broken into three seperate PRs. A core one for most of the changes, one for things related to meters, and one for this outline/tooltip stuff

code/_helpers/client.dm
DeadLineSS13: > this proc name is less descriptive, keep the old name
> 
> and why fewer vars? You've removed some functionality here
> It took a LOT of work to get this thing working correctly, i'm suspicious of changes

Now view is only string var so example "4x4" and etc. Because name proc changed. view_radius <=> temp_view (small remaning).

code/_helpers/client.dm
DeadLineSS13: > what is this?

Auto fit game place. Removes black edges automatically.

code/_helpers/client.dm
DeadLineSS13: > this temp view var was necessary, do you understand what it was used for before you remove it? How have you resolved its problems?

Re-read the view_radius code and you will understand that they are the same thing.

code/_onclick/click_catcher.dm
DeadLineSS13: > void is not a descriptive name, call it click_catcher
> 
> Remaking it here was necessary to avoid view size problems, how have you resolved that?

UpdateGreed solve this problem.

code/__defines/misc.dm
DeadLineSS13: > this seems bad, these should be calculated from client's viewsize not hardcoded

Deleted

code/_onclick/hud/hud.dm
DeadLineSS13: Updated

code/_onclick/click_catcher.dm
DeadLineSS13: > why have you removed all this documentation and helper procs?


I believe that the documentation is outdated for this code and in addition it is not needed. Since here she performs the most basic mechanics. It is generally better not to make changes in it.

code/_onclick/click_catcher.dm
DeadLineSS13: > greed? this needs a better name and explanation

Greed - We took max place View.

code/_onclick/hud/fullscreen.dm
DeadLineSS13: > why are you doing this here? it should be inside the loop


It is not necessary to contain it in a loop. And it seems like it does not reassign a memory cell every time. Not noticeable, but a little nice.

code/_onclick/hud/fullscreen.dm
DeadLineSS13: > no abbreviations please, give this variable a full title descriptive of what it does

fullscreen_view = 480x480 (15x15 tiles)

code/_onclick/hud/human.dm
DeadLineSS13: Updated

code/_onclick/hud/meter/meter.dm
DeadLineSS13: > cleaning itself up from the client screen makes sense


It's better to never work directly with such important things.

code/_onclick/hud/meter/meter.dm
DeadLineSS13: > this proc was definitely needed

We don't need custom proc.

code/_onclick/hud/meter/meter.dm
DeadLineSS13: > so was this, i think most of the changes to meters will need reverted

Nope

code/game/atoms.dm
DeadLineSS13: > We were just arguing about memory footprint this morning. Adding a new list to all atoms is a terrible idea.
> 
> Extensions can manage filters for the few things that use them, no new list on atoms

Reverted

code/game/objects/_items.dm
DeadLineSS13: > string form of force? You mean the number that indicates how much damage it does? Why would you cache that

Reverted

code/game/objects/_items.dm
DeadLineSS13: > use an extension for managing cooldowns and last checks if such is necessary

Reverted

code/game/objects/_items.dm
DeadLineSS13: > this needs explanation

Reverted

code/game/objects/_items.dm
DeadLineSS13: > this definitely doesn't need to be cached, just run the proc to fetch it on demand
> 
> and don't hardcode numbers here, use the same defines that weapon forces are derived from

Reverted

code/game/objects/_items.dm
DeadLineSS13: > this puts an outline on anything someone hovers over? That seems like it would have a huge performance cost
> 
> is it restricted to things in range?

Reverted

code/game/objects/_items.dm
DeadLineSS13: > this would be better accessed with a click handler that is only applied to conscious humans so that ghosts don't get this far in the first place

Reverted

code/modules/client/client_defines.dm
DeadLineSS13: > how does this differ from view? Doesn't it do the same thing temp_view did?

view_radius <=> temp_view

code/modules/client/client_defines.dm
DeadLineSS13: > why are tooltips cached in a list? Shouldn't they just be part of screen generally?

Because this show only for clients. And all mobs have hud.

code/modules/client/client_procs.dm
DeadLineSS13: > i dont see a need to seperate generate and apply here, we wouldn't want a click catcher that isn't sized properly, right?

It works fine thanks to the Upgradegreed.

code/modules/client/client_procs.dm
DeadLineSS13: > QDEL_NULL_LIST

No lists there

code/modules/client/preferences.dm
DeadLineSS13: > if this feature is going to be added it should not be optional, people would turn it off to get a gameplay advantage so nobody can see where they're looking

reverted

code/modules/client/verbs/ooc.dm
DeadLineSS13: > taking guesses in a loop seems hella sloppy, why is it doing this? Surely the right values can be calculated?


Byond may not be able to figure it out exactly.

code/_onclick/click_catcher.dm
DeadLineSS13: > these require explanation, maximum safe?

To prevent sprites from stretching.

code/modules/client/client_procs.dm
DeadLineSS13: > if (!QDELETED(src))

Ready

code/modules/mob/inventory.dm
DeadLineSS13: > what's this new parameter used for? i don't think it belongs here

Deleted

code/modules/mob/mob_defines.dm
DeadLineSS13: > rename this define WORLD_VIEW_RANGE

Renamed

code/_onclick/hud/fullscreen.dm
NanakoAC: don't tell me, change its name to that!

NanakoAC (approved):

PR #1727, opened 2021-09-23
Stamina bar fix

DTraitor:

### DIFF
--- a/code/modules/client/ui_style.dm
+++ b/code/modules/client/ui_style.dm
@@ -34,7 +34,7 @@
 	if(!UI_style_color_new) return
 
 	//update UI
-	var/list/icons = usr.hud_used.adding + usr.hud_used.other + usr.hud_used.hotkeybuttons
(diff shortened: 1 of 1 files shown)


code/modules/client/ui_style.dm
```diff
@@ -34,7 +34,7 @@
 	if(!UI_style_color_new) return
 
 	//update UI
-	var/list/icons = usr.hud_used.adding + usr.hud_used.other + usr.hud_used.hotkeybuttons
+	var/list/icons = usr.hud_used.adding + usr.hud_used.other + usr.hud_used.hotkeybuttons - usr.hud_used.stamina_bar
```
NanakoAC, line 37: what exactly is being fixed here, and why does subtracting this specific icon do things?

this doesn't seem like the best way to do it

DTraitor: If you'll look at code below you'll see it sets custom icon, while stamina bar uses it's own file

NanakoAC (approved):

PR #1736, opened 2021-09-24
A lot less cybernetic fun.

Ketrai:
Stops humans from picking artificial organs in the character setup menu. Artificial limbs can still be picked.

This change is intended to stop humans from avoiding the negative effects from drugs, and the fact it's unexplainable from an IC standpoint why someone got an expensive cybernetic organ, instead of a replacement organ from the BPL. They also don't really serve a good purpose compared to normal prosthetics, which are handy when the BPL isn't quite ready yet.

### DIFF
--- a/code/modules/client/preference_setup/general/03_body.dm
+++ b/code/modules/client/preference_setup/general/03_body.dm
@@ -600,7 +600,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 			if("Kidneys")
 				organ = BP_KIDNEYS
 
-		var/list/organ_choices = list("Normal","Assisted","Synthetic")
+		var/list/organ_choices = list("Normal")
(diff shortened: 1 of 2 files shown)


code/modules/client/preference_setup/general/03_body.dm
```diff
@@ -600,7 +600,7 @@ var/global/list/valid_bloodtypes = list("A+", "A-", "B+", "B-", "AB+", "AB-", "O
 			if("Kidneys")
 				organ = BP_KIDNEYS
 
-		var/list/organ_choices = list("Normal","Assisted","Synthetic")
+		var/list/organ_choices = list("Normal")
```
NanakoAC, line 603: seems like assisted ones should be allowed

Ketrai: Assisted ones may not properly process chems, is the issue.

PR #1738, opened 2021-09-24
Nerfs leaper stun duration

Ketrai:
Lowers leaper stun duration from 5 seconds to 3. This is a decrease of 40%.

I considered doing a smaller nerf to the stun at first, but I'd rather not lower leap cooldown any further for mobility reasons, and this makes it harder to spam stuns together with another leaper.


NanakoAC:
leapers do need some work but not this, jumping on people and pinning them down is intended, what needs fixing is the knocking them back, and the being able to leap from close range. I'll handle this myself when new leaper sprites are ready

PR #1740, opened 2021-09-24
Lowers marker initial biomass.

Ketrai:
Lowers the amount of biomass the marker starts with from 250 to 200.
Marker is still expected to have about 400 biomass as it activates due to passive gain. This should make it slightly more difficult to expand aggressively early-on, and make early tier 2 a lot more difficult.


NanakoAC (approved):
this is reverting a change that was made months ago, but i guess things change

PR #1743, opened 2021-09-24
Small update Processing code

DeadLineSS13:
## About the PR
Small changes in Processing code. Now we don't use is_processing, flags better. (No lazy code in processing please)


NanakoAC (changes requested):
We JUST had work done on this. 
is_processing isnt just a boolean, it contains a reference to the processing subsystem being used, this is important.

changing it into a binary flag is a downgrade

NanakoAC:
following discussion in dev closet, we won't be using this PR

PR #1756, opened 2021-09-25
Fixes buckled mobs by JavelinGun and add Trait system

DeadLineSS13:
## About ther PR
Fixed buckled mobs by JavelinGun.
Added Trait system. (Now we can reduce the number of changes that contain only TRUE, FALSE or Bitflags.)

## Changelog
🆑 DeadLineSS13
fix: Fixed buckling for javelin gun.
add: Added Trait system. (This thing for DEVS)
/🆑


code/modules/mob/living/living.dm
```diff
@@ -382,12 +382,29 @@ default behaviour is:
...
 
 	//unbuckling yourself
-	if(buckled && !HAS_TRAIT(src, TRAIT_BUCKLED))
-		spawn() escape_buckle()
-		return TRUE
+	if(buckled)
+		if(HAS_TRAIT(src, TRAIT_BUCKLED))
+			var/list/sources = status_traits[TRAIT_BUCKLED]
+			for(var/i in sources)
+				if(istype(i, /obj/item/javelin))
```
NanakoAC: aghh noo this is not what i said. Don't put specific item paths in here

resist_buckle is already called on the buckled item, just override that in the javelin, and redirect attack_hand to it too

NanakoAC (changes requested):
incorrect javelin solution

code/modules/mob/living/living.dm
```diff
@@ -389,14 +389,8 @@ default behaviour is:
 		if(HAS_TRAIT(src, TRAIT_BUCKLED))
 			var/list/sources = status_traits[TRAIT_BUCKLED]
 			for(var/i in sources)
-				if(istype(i, /obj/item/javelin))
-					var/obj/item/javelin/j = i
-					j.attack_hand(src)
-				else
-					if (buckled.resist_buckle(src))
```
NanakoAC: still no, this part was fine before. It calls resist for the process of escaping, then escape to finalise. Restore this section

in the javelin, call resist buckle and escape buckle in attack hand

code/game/objects/buckling.dm
```diff
@@ -22,13 +22,7 @@
 	return ..()
 
 /obj/proc/escape_buckle(mob/user, support = FALSE)
```
NanakoAC, line 24: we should have two procs, 

resist_buckle should do any special stuff needed to deal with the buckling, like do-after times. And it should return true or false.

escape_buckle should handle actually getting the mob out from the buckle and letting them go

call these two in resist code like `if (obj.resist_buckle(user)) obj.escape_buckle(user)`

PR #1784, opened 2021-09-28
Hyperzine & floors

Ketrai:
- Disables the recipe for hyperzine.
Hyperzine can still be found in maintenance as zoom pill bottles and anabolic steroids. It should just be a very rare 'treat' now, which serves it right granted how powerful it is currently.

- Removes floor digging by hand.
We already removed this from necromorphs for the very same issue as necros digging through levels or making holes to curb survivor movement. You can still remove floors by using the proper tools. Which takes longer, requires engineering skill, and the appropriate inventory space allocated to tools.

### DIFF
--- a/code/game/turfs/simulated/floor_attackby.dm
+++ b/code/game/turfs/simulated/floor_attackby.dm
@@ -15,6 +15,7 @@
 	if (flooring && flooring.attackby(I, user, src))
 		return TRUE
 
+/* disabled until further notice. This isn't intended, use tools instead.
 	//Attempting to damage floors with things
 	//This has a lot of potential to break things, so it's limited to harm intent.
 	//This supercedes all construction, deconstruction and similar actions. So change your intent out of harm if you don't want to smack the floor
@@ -36,6 +37,7 @@
 				visible_message(SPAN_DANGER("[user] ineffectually hits [src] with [I]"))
 			user.set_click_cooldown(DEFAULT_ATTACK_COOLDOWN*1.5) //This longer cooldown helps promote skill in melee combat by punishing misclicks a bit
 			return TRUE
+*/
 
 	for(var/atom/movable/A in src)
 		if(A.preventsTurfInteractions())
--- a/code/modules/reagents/Chemistry-Recipes.dm
+++ b/code/modules/reagents/Chemistry-Recipes.dm
@@ -259,11 +259,13 @@
 	inhibitors = list(/datum/reagent/sugar = 1) // Messes up with inaprovaline
 	result_amount = 2
 
+/* disabled for now. You can still spawn in anabolic steroids for events, but this needs to be rebalanced.
 /datum/chemical_reaction/hyperzine
 	name = "Hyperzine"
 	result = /datum/reagent/hyperzine
 	required_reagents = list(/datum/reagent/sugar = 1, /datum/reagent/phosphorus = 1, /datum/reagent/sulfur = 1)
 	result_amount = 3
+*/
 
 /datum/chemical_reaction/cryoxadone
 	name = "Cryoxadone"


NanakoAC:
I don't like this

Disabling floor attacking like that is a core functionality, we can't do that. Its shared behaviour across all floors including things like dirt, and in future, snow. If digging through ship floors is a problem, change values on the underplating instead

Hyperzine is already on my radar. Rebalance, don't remove.
The plan is to remove the normal movespeed bonus, and incorporate the encumbrance reduction into the new encumbrance system.

You could do that by adding a new CE_XXX effect and checking for it in encumbrance update code

Ketrai:
> I don't like this
> 
> Disabling floor attacking like that is a core functionality, we can't do that. Its shared behaviour across all floors including things like dirt, and in future, snow. If digging through ship floors is a problem, change values on the underplating instead

Dirt and snow actually have their own checks for specific tools. The part I'm disabling only allows you to directly attack floors that have health and can be destroyed. As I have explained nanako, this gets abused religilously. It completely takes away the tension from a fight when survivors can dig a hole and escape the horde in a really small timeframe. Rather than holding out or making a hail mary push for the pods. It takes some necromorphs longer to even open a normal door than the time it takes to circumvent huge stretches of the playable area.

Digging a wall lets you traverse a level more easily. Digging the floor lets you invalidate map design. Also, it trashes the engineer profession. The less you're able to do as a single survivor. The more cooperation is encouraged. It's also kinda immersion breaking too. We're not trying to be some kind of action genre where you shoot up the floor or smash it with a blade. Hell, you could even dig through the floor by slapping it with a spoon enough if you have the patience. 

Either an exception needs to be made for all common spaceship flooring. Or what is easier in my  books, you enable a proc for individual floors you do want to get destroyed. Rather than making this the default behavior for all floors universally.

Ketrai:
> Hyperzine is already on my radar. Rebalance, don't remove. The plan is to remove the normal movespeed bonus, and incorporate the encumbrance reduction into the new encumbrance system.
> 
> You could do that by adding a new CE_XXX effect and checking for it in encumbrance update code

It's not removing hyperzine though. As said, this just makes it much rarer. Which is a good bandaid solution until you get time to rebalance it. It commonly ruins rounds for survivors and necromorphs alike because some people notoriously abuse the effect. That's not something you leave lying around.

Also just making it target encumberance is also a bad idea. The vintage rig is meant to be slow for example. If you can just circumnavigate the effect, why bother with other rigs?

NanakoAC:
> It's not removing hyperzine though. As said, this just makes it much rarer. Which is a good bandaid solution until you get time to rebalance it. It commonly ruins rounds for survivors and necromorphs alike because some people notoriously abuse the effect. That's not something you leave lying around.
> 
> Also just making it target encumberance is also a bad idea. The vintage rig is meant to be slow for example. If you can just circumnavigate the effect, why bother with other rigs?

this is effectively removing it. More importantly, its removing it from medical, and thus taking away gameplay from an already underdeveloped area. This band aid solution makes medical worse in the short term

Making it target encumbrance is in line with its original design intent, and gives it a specific purpose rather than being a no-brainer speedboost. It won't be such a powerful effect that it completely invalidates rigs, just negating a single point of encumbrance

Its not hard to do, i can get to it soon, or guide you through doing it

NanakoAC:
> 
> 
> > I don't like this
> > Disabling floor attacking like that is a core functionality, we can't do that. Its shared behaviour across all floors including things like dirt, and in future, snow. If digging through ship floors is a problem, change values on the underplating instead
> 
> Dirt and snow actually have their own checks for specific tools. The part I'm disabling only allows you to directly attack floors that have health and can be destroyed. As I have explained nanako, this gets abused religilously. It completely takes away the tension from a fight when survivors can dig a hole and escape the horde in a really small timeframe. Rather than holding out or making a hail mary push for the pods. It takes some necromorphs longer to even open a normal door than the time it takes to circumvent huge stretches of the playable area.
> 
> Digging a wall lets you traverse a level more easily. Digging the floor lets you invalidate map design. Also, it trashes the engineer profession. The less you're able to do as a single survivor. The more cooperation is encouraged. It's also kinda immersion breaking too. We're not trying to be some kind of action genre where you shoot up the floor or smash it with a blade. Hell, you could even dig through the floor by slapping it with a spoon enough if you have the patience.
> 
> Either an exception needs to be made for all common spaceship flooring. Or what is easier in my books, you enable a proc for individual floors you do want to get destroyed. Rather than making this the default behavior for all floors universally.

add extra resistance to the underplating and no more ship digging, problem solved. Give it a high enough value that most hand weapons can't damage it

Its best not to delete core functionality to deal with specific case issues

Ketrai:
I'll work on another solution for this. Most likely just disabling the attackby for a certain floor type specifically...

NanakoAC:
Made obsolete by linked PR above

PR #1793, opened 2021-09-30
Fixes part issues Assets

DeadLineSS13:
## About the PR
Fixed part issue Assets NanoUI.

## Changelog
🆑 DeadLineSS13
refactor: Spritesheet craft.
🆑


code/datums/craft/menu.dm
```diff
@@ -51,9 +51,11 @@
 	data["cur_item"] = null
 
 	if(CR)
+		var/datum/asset/spritesheet/sheet = get_asset_datum(/datum/asset/spritesheet/craft)
+		var/list/result_item = get_craft_item(CR.result)
```
Kmc2000, line 55: Consider a check for this list existing here?

code/modules/nano/nanoui.dm
```diff
@@ -253,8 +263,9 @@ nanoui is used to open and update nano browser uis
   *
   * @return nothing
   */
-/datum/nanoui/proc/add_script(file)
-	scripts.Add(file)
+/datum/nanoui/proc/add_script(name, file)
+	scripts["[ckey(name)].js"] = file
```
Kmc2000, line 267: This seems weird, could you please explain why you’ve made this change?
Does every client need their own version of a js file?

code/datums/craft/menu.dm
DeadLineSS13: > Consider a check for this list existing here?

It will exist with 100% probability, so no checks are needed.

code/modules/nano/nanoui.dm
DeadLineSS13: https://github.com/tgstation/tgstation/pull/20890 old issue

Kmc2000 (approved):
Looks sound, uses tried & tested code (per that old PR)

NanakoAC:
there's a linter check in the red

NanakoAC (approved):
looks like the linter issue here is the one i fixed. Please update this and it can be merged

ghost:
@DeadLineSS13 Please resolve conflicts.

PR #1797, opened 2021-09-30
Small upgrade Observation

DeadLineSS13:
## About the PR
RAISE_EVENT worked faster then GLOB.thing.raise_event. And RAISE_EVENT can return value.

## Changelog
🆑 DeadLineSS13
refactor: Observation small refactored.
/🆑


code/datums/datum.dm
```diff
@@ -9,6 +9,15 @@
 	/// Datum level flags
 	var/datum_flags = NONE
 
+	/**
+	  * Any datum registered to receive observations from this datum is in this list
+	  *
+	  * Lazy associated list in the structure of `observation:registree/list of registrees`
+	  */
+	var/list/observations
```
Kmc2000, line 17: You have an uninitialised list here, I’m assuming it’s initialised elsewhere?

May be useful to include a wrapper to do this easily as you do with RaiseEvent

code/datums/datum.dm
```diff
@@ -9,6 +9,15 @@
...
+	/// Lazy associated list in the structure of `observations:proctype` that are run when the datum receives that observation
+	var/list/list/datum/callback/observation_procs
```
Kmc2000, line 19: Strictly typing this list may lead to wackiness, I’m not 100% sure on how well BYOND actually deals with these things. I believe the convention is to use a loosely typed assoc list with type checking as needed.

Kmc2000: Again, if not already existent, a wrapper proc to add an event listener could be nice here.

Kmc2000 (approved):
Pretty clever way of doing this. Seems to be a very stripped down version of signals, which I’m not sure bay has. In that case though, consider porting them over, it’s basically this but somewhat easier to use.

I’d say this looks fine at a glance, though I’m not 100% familiar with this codebase’s way of handling events.

DeadLineSS13:
> Pretty clever way of doing this. Seems to be a very stripped down version of signals, which I’m not sure bay has. In that case though, consider porting them over, it’s basically this but somewhat easier to use.
> 
> 
> 
> I’d say this looks fine at a glance, though I’m not 100% familiar with this codebase’s way of handling events.

Observation is just a version of Signals Bay. The problem is that all the events were stored globally and the raise_event response was much longer in the end. I decided to attach this moment to datums (that is, as on TG). But at the same time, it is much better than Signals TG, since data on what events there are can be viewed through GLOB. And at the same time, you can change the register/unregister/raise_event for each event. That is, it is more flexible and more lucid than Signals.

code/datums/datum.dm
DeadLineSS13: Byond have dynamic typization.
Example create var/thing = list()
thing will contain list, but don't have memory access for work with list (thing.len, thing.Copy, thing.Cut and etc)

PR #1802, opened 2021-10-01
Puker and Spitter projectiles are now slower

DTraitor:
### Purpose
After we got a new server it became really hard to catch spitter projectiles due to ping and their high speed. This PR makes them slower


Ketrai:
These are some pretty significant nerfs. This also affects side-stepping projectiles. This necromorph is already in a bad spot in the current meta.

NanakoAC:
applying smooth movement to projectiles would be a better solution i think
maybe a speed reduction is warranted, but this degree of change seems too much, incremental changes

DTraitor:
Closing for now

PR #1832, opened 2021-10-04
Peripheral Items

Pebblesthe2nd:
The old Bay sprites aren't my favourites, adding a few of mine that I've made in ages past.
First some various items.
Camera & films (cause the old ones are not good and huge for very marginal items rarely used.)
Defib (less oversized and gameplay around it on tables etc better thanks to this)
IV (just recolored darker, more contrast)

https://imgur.com/a/qMYTIoM


Ketrai:
image
Please copy paste images into github like this to get an imbedded upload rather than linking to third party image hosting sites.

Ketrai:
Generally in my books items that find themselves in the game world often and are less important can be smaller. (Not to mention these sprites generally feature better shading which is something we aim for)

Otherwise I don't mind these. I'd just prefer new sprites for existing items to be deadspace themed.

Pebblesthe2nd:
Interesting, I was not aware such a convenient function for images was included in Git. Thank you for the tip.
Re: the stuff not being specifically DS themed, I understand the desire, but at the same time there are many hundreds of placeholder sprites and its likely to remain so for a very long time and upgrading the worst offenders has no real downside as I see it.

Not to look a gift horse in the mouth and all that does apply as well =)

Snypehunter007:
My two cents is that most of them look fine, though I think the defibs are maybe a touch too small.

PR #1833, opened 2021-10-04
Fashion upgrade

Pebblesthe2nd:
Some clothing items in dire need of love gets some. Engineering/Medical.
Affected clothing
Gloves : Nitrile, Latex, Insulated
Suits : Rad suit, Bomb suit (only icons, very minor onmob edit)
Masks: Sterile
Uniforms: Hazard jumpsuit (only some added depth of shading and better icon)

what it looks like
https://imgur.com/a/ypoO8zL


NanakoAC:
looks good to me, although wasnt the old nitrile glove sprite more accurate? (it definitely is ugly, but so are the real ones)

Ketrai:
Same notes as #1832 really.

Pebblesthe2nd:
Re: accuracy, nitrile gloves vary quite a bit in color depending on your nation and employer. I'm more used to white or darker blues than the very bright ones, but I've seen those too on occasion, so basically from a realism viewpoint either works. The current shade is palette adjusted to fit the other surgery scrub stuff and hairnets I made, but those might be redundant, no clue what the specific visual concepts are for medical long term. Either way they need gloves I figure =)

Personal preference wise I obviously like the darker more but if someone goes and adjusts the color temp slider and makes it a little more neon its fine by me, as long as they keep the shading. Unshaded sprites crushes my soul.

Snypehunter007:
Personal thoughts, blue gloves look kinda meh.

Sterile mask is more distinctive in blue and stands out more (in a good identification wise way), the proposed white version is kinda meh.

If you're going to do the rad suit, I'd recommend doing the security rad suit as well, since it is basically the same thing just with a red stripe.

Bomb suit . . . I mean, it isn't bad, definitely an improvement. We just aren't going to use it, most likely?

PR #1842, opened 2021-10-05
Fixes log_runtime moment.

DeadLineSS13:


code/_helpers/unsorted.dm
```diff
@@ -1070,6 +1070,15 @@ var/list/WALLITEMS = list(
 /proc/crash_with(msg)
 	CRASH(msg)
 
+GLOBAL_REAL_VAR(list/stack_trace_storage)
+/proc/gib_stack_trace()
+	stack_trace_storage = list()
```
Kmc2000, line 1075: This looks like an unmanaged globalvar and not a GLOB managed one

code/controllers/configuration/entries/general.dm
```diff
@@ -48,9 +48,6 @@ Basics, the most important.
 /datum/config_entry/flag/log_hrefs  //logs all links clicked in-game. Could be used for debugging and tracking down exploits
     protection = CONFIG_ENTRY_LOCKED
 
-/datum/config_entry/flag/log_runtime    //logs world.log to a file
-	config_entry_value = FALSE
```
Kmc2000, line 52: This isn’t ideal, as it’s now a hard define. Though ds13 doesn’t really worry about forks so I guess it works.

code/game/world.dm
```diff
@@ -126,24 +125,29 @@ GLOBAL_VAR(restart_counter)
...
 		GLOB.using_map.perform_map_generation()
 
-	if(CONFIG_GET(flag/log_runtime))
-		var/runtime_log = file("data/logs/runtime/[date_string]_[time2text(world.timeofday, "hh:mm")]_[GLOB.round_id].log")
-		runtime_log << "Game [GLOB.round_id] starting up at [time2text(world.timeofday, "hh:mm.ss")]"
-		log = runtime_log
-
 	update_status()
 
-	GLOB.timezoneOffset = text2num(time2text(0,"hh")) * 36000
```
Kmc2000, line 146: Can you explain this change briefly?

Kmc2000 (commented):
Couple points of interest, looks sound though

NanakoAC:
please address the comments above

code/game/world.dm
DeadLineSS13: > Can you explain this change briefly?

Move code. Because in future we can add No Inits params.

code/controllers/configuration/entries/general.dm
DeadLineSS13: > This isn’t ideal, as it’s now a hard define. Though ds13 doesn’t really worry about forks so I guess it works.

Old system logging runtime. It was logged incorrectly and if this flag was enabled, it was simply duplicated in 2 logs.

code/_helpers/unsorted.dm
DeadLineSS13: > This looks like an unmanaged globalvar and not a GLOB managed one

We don’t need to manage this var. Because this dangerous.

NanakoAC (approved):

PR #1843, opened 2021-10-05
Minor loadout clothing addition

Pebblesthe2nd:
Four pieces of clothing for loadouts for a little added customisation.

For MD:s : EMS jacket
Engineer: Hazard overall
Cargo people: Bomber jacket
Civilians: Leather jacket

Just a bit of optional color for characters that doesn't go to the bar with their full RIG deployed. Adjusted their descriptions a little.

### DIFF
--- a/code/modules/client/preference_setup/loadout/lists/clothing.dm
+++ b/code/modules/client/preference_setup/loadout/lists/clothing.dm
@@ -8,3 +8,41 @@
 	display_name = "scarf"
 	path = /obj/item/clothing/accessory/scarf
 	flags = GEAR_HAS_COLOR_SELECTION
+
+/datum/gear/clothing/hazard
+	display_name = "hazard overalls"
+	slot = slot_w_uniform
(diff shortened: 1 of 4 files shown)


Ketrai:
Again, would really like proper screenshots in these.

Snypehunter007:
> 
> 
> Again, would really like proper screenshots in these.

He added them to the loadout, he didn't replace the icons this time.

That said, personally, going to go with a negative for the proposed changes.

Pebblesthe2nd:
Sprite-0001
Looks like this.
It makes no sense to me that the hazard jumpsuit is almost civilian only rather than engineers since it only spawn in maints for some strange reason. As for the others, just minor flavour items, having the chef have a optional jacket to cover his Mario getup when on break doesn't seem particularly offensive to me.
Also trivial to edit should something become overused or problematic.

Ketrai:
As much as I like these we're aiming to reduce the amount of non dead space clothing in the loadout down the line. Not increase it.

PR #1849, opened 2021-10-07
Psychic Tracer Runtime

### LINKED ISSUE #1819: Psychic Tracer null visualnet tiles
[12:50:17] Runtime in psychic_tracer.dm, line 51: Cannot execute null.get visualnet tiles().
proc name: get visualnet tiles (/obj/effect/psychic_tracer/get_visualnet_tiles)
src: the psychic tracer (/obj/effect/psychic_tracer)
src.loc: null
call stack:
the psychic tracer (/obj/effect/psychic_tracer): get visualnet tiles(/datum/visualnet/necrovision (/datum/visualnet/necrovision))
/datum/chunk/necrovision (/datum/chunk/necrovision): get datum visible turfs(the psychic tracer (/obj/effect/psychic_tracer))
/datum/chunk/necrovision (/datum/chunk/necrovision): acquire visible turfs(/list (/list))
/datum/chunk/necrovision (/datum/chunk/necrovision): update(0)
Chunks (/datum/controller/subsystem/chunk): fire(0)
Chunks (/datum/controller/subsystem/chunk): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop()
Master (/datum/controller/master): StartProcessing(0)

NanakoAC:
Fixes #1819

### DIFF
--- a/code/modules/mob/dead/observer/freelook/marker/necrovision.dm
+++ b/code/modules/mob/dead/observer/freelook/marker/necrovision.dm
@@ -19,7 +19,7 @@
 			continue
 
 
-		var/list/visible = get_datum_visible_turfs(source)
+		var/list/visible_turfs = get_datum_visible_turfs(source)
 
 		//Special return value to indicate that we are invalid and should be removed
 		if (visible == PROCESS_KILL)
--- a/code/modules/mob/dead/observer/freelook/marker/necrovision.dm
+++ b/code/modules/mob/dead/observer/freelook/marker/necrovision.dm
@@ -17,7 +17,16 @@
 		if (!istype(source))
 			sources -= source
 			continue
-		for(var/t in get_datum_visible_turfs(source))//source.get_visualnet_tiles(visualnet))
+
+
+		var/list/visible = get_datum_visible_turfs(source)
+
+		//Special return value to indicate that we are invalid and should be removed
+		if (visible == PROCESS_KILL)
+			sources -= source
+			continue
+
+		for(var/t in visible)//source.get_visualnet_tiles(visualnet))
 			visible[t] = t
 
 
--- a/code/modules/mob/dead/observer/freelook/marker/necrovision.dm
+++ b/code/modules/mob/dead/observer/freelook/marker/necrovision.dm
@@ -22,11 +22,11 @@
 		var/list/visible_turfs = get_datum_visible_turfs(source)
 
 		//Special return value to indicate that we are invalid and should be removed
-		if (visible == PROCESS_KILL)
+		if (visible_turfs == PROCESS_KILL)
 			sources -= source
 			continue
 
-		for(var/t in visible)//source.get_visualnet_tiles(visualnet))
+		for(var/t in visible_turfs)//source.get_visualnet_tiles(visualnet))
 			visible[t] = t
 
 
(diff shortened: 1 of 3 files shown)


code/modules/mob/dead/observer/freelook/marker/necrovision.dm
```diff
@@ -17,7 +17,16 @@
 		if (!istype(source))
 			sources -= source
 			continue
-		for(var/t in get_datum_visible_turfs(source))//source.get_visualnet_tiles(visualnet))
+
+
+		var/list/visible = get_datum_visible_turfs(source)
```
Kmc2000: Return an empty list instead? null perhaps? this works, but it’s pretty weird

code/modules/mob/dead/observer/freelook/marker/signal_abilities/psychic_tracer.dm
```diff
@@ -48,9 +48,19 @@
 
 
 /obj/effect/psychic_tracer/get_visualnet_tiles(var/datum/visualnet/network)
+
+	//Unsure why this is happening, but it seems that we aren't getting properly removed from chunk lists. Return PROCESS_KILL to ensure that happens
+	if (!EM)
+		if (!QDELETED(src))
```
Kmc2000, line 54: Why not a normal qdel()? That’s an extra timer needing to exist for each time this goes wrong, one to consider

Kmc2000 (approved):
Looks good overall, few ideas. Generally I’d prefer to hunt down the cause of the issue rather than patch over it but this is small enough

NanakoAC:
> 
> 
> Looks good overall, few ideas. Generally I’d prefer to hunt down the cause of the issue rather than patch over it but this is small enough

I believe i know the cause, but it's not easy to fix.

When sources are removed from a visualnet, they are also removed from the sources list of all chunks within range. That is, within range of where they are at that moment

in the case of a mob suddenly getting deleted, its got a null location to search from.
I reckon also that there are cases of something being deleted while moving, and the chunks holding onto references are ones that had just gone out of range and thus the thing doesnt get removed from them

that second one at least, isn't easy to fix without slowing down the visualnet system as a whole

code/modules/mob/dead/observer/freelook/marker/necrovision.dm
NanakoAC: returning an empty list is valid data, can't use that for an error code

code/modules/mob/dead/observer/freelook/marker/signal_abilities/psychic_tracer.dm
NanakoAC: the thing is going to try to remove itself when deleted, and its going to be removed with returning this error code anyways, it felt like having all that in one frame might be redundant, and lead to possible race conditions or infinite loops

but you're right that this could cause lots of timers if there are many chunks holding onto one reference
what do you reckon, should i change it?

code/modules/mob/dead/observer/freelook/marker/necrovision.dm
DTraitor: use list.len then

code/modules/mob/dead/observer/freelook/marker/signal_abilities/psychic_tracer.dm
Kmc2000: I think a straight qdel here honestly, QDEL_IN wouldn’t really solve the race condition. Consider refactoring this check onto whatever calls the visualnet to avoid modifying the list on the fly / at random frames?

Could also just remove it from the managed list here manually and let the GC take care of it.

PR #1853, opened 2021-10-08
Balance update V3

Ketrai:
The primary focus of this update is to lower the overall power level of unitologists, and the chaos they tend to cause. They're a bit too disruptive to the round currently, and the ease of access to deadly equipment doesn't help.
 
 The second part is necromorph nerfs. More specifically, spread. If anything, corruption can claim medbay within 5-15 minutes, depending on how adamant necros are. And no place is safe from dedicated spread. This should slow them down a bit, and give people more time to respond to the infection.

 - Rebalanced antag equipment.
  - Uni zealot rig cost increased. (Now you can't buy a rig, a weapon, and three mags anymore. Choose either a rig and a divet, just the rig, or just a bullpup instead)
  - Earthgov agents can now get a rig as well. (Primarily for parity. I'll make a unique rig sprite for this later)
  - Zealot smg replaced by bullpup, cost increased. (Because using a classic ballistic smg is silly.)
  - Implant cost increased. (you were able to get 2. Unis could effectively enslave half the crew on some rounds)
  - c4 cost increased. (You can now buy up to a maximum of 5. Down from 7. This will lower the amount of chaos caused by mass c4 detonations)
  - zealots now have access to the divet. (Simple sidearm to use with the rig. They can't buy additional ammo, it's really just a backup.)
  - Arm/hand health increased further. (It's mostly incredibly unfun for humans to loose the means to fight effectively. Necros can still easily kill people with infector aid, or by damaging them enough.)
  - Marker can now spawn in the undertram. (2 spawns, 2/5 total spawns. Bit Experimental. They get access to some vending machines right away, but are much further from topside bio objectives. Players have been complaining about medical falling really early, making it hard to effectively treat wounds even when doctors are still alive.)
  - Branch cost decreased. (underused node. This makes it more attractive. 10% decrease)
  - Root distance and spread speed nerfed significantly. (12 tiles now. This is still a lot of absolute distance. But also makes it easier to spread a given direction. This makes propagators and branches more attractive again, instead of signals only spawning roots 33% decrease in speed)
  - Propagator spread speed increased modestly. (Makes it more attractive. 20% increase)

### DIFF
--- a/code/modules/necromorph/corruption/growth.dm
+++ b/code/modules/necromorph/corruption/growth.dm
@@ -7,7 +7,7 @@
 	density = FALSE
 
 	var/range = 14
-	var/speed = 1.25
+	var/speed = 1.5
 	var/falloff = 0.1
 	var/limit = null	//Maximum number of tiles it can support
 
@@ -66,7 +66,7 @@
 	name = "branch"
 	id = "branch"
 	desc = ""
-	energy_cost = 150
+	energy_cost = 135
 	placement_atom = /obj/structure/corruption_node/growth/branch
 
 
@@ -82,8 +82,8 @@
 	density = FALSE
 	marker_spawnable = FALSE
 
-	range = 40
-	speed = 3
+	range = 12
+	speed = 2
 	falloff = 0.025
 	limit = 60
 	randpixel = 4
--- a/code/datums/uplink/ammunition.dm
+++ b/code/datums/uplink/ammunition.dm
@@ -61,7 +61,6 @@
 /datum/uplink_item/item/ammo/c45uzi/special
 	item_cost = 2
 	is_special = TRUE
-	antag_roles = list(MODE_UNITOLOGIST, MODE_UNITOLOGIST_SHARD)
 
 /datum/uplink_item/item/ammo/a10mm
 	name = "10mm SMG Magazine"
@@ -70,6 +69,11 @@
 
 /datum/uplink_item/item/ammo/a10mm/special
 	item_cost = 2
+	is_special = TRUE
+
+/datum/uplink_item/item/ammo/bullpup/special
+	item_cost = 8
+	path = /obj/item/ammo_magazine/bullpup
 	is_special = TRUE
 	antag_roles = list(MODE_UNITOLOGIST, MODE_UNITOLOGIST_SHARD)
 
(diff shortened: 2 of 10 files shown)


code/modules/necromorph/corruption/growth.dm
```diff
@@ -7,7 +7,7 @@
 	density = FALSE
 
 	var/range = 14
-	var/speed = 1.25
+	var/speed = 1.5
```
NanakoAC, line 10: These corruption changes are mostly fine for now, but they will likely be overwritten when i overhaul corruption spread in future

code/modules/necromorph/corruption/growth.dm
```diff
@@ -82,8 +82,8 @@
 	density = FALSE
 	marker_spawnable = FALSE
 
-	range = 40
-	speed = 3
+	range = 12
+	speed = 2
```
NanakoAC, line 86: this is a crazy degree of change to do all at once, you have no idea how this is going to affect things. 
Go with a smaller change, like to 30 to start with and see how it goes

NanakoAC (approved):

NanakoAC (changes requested):
the root change is too much

code/modules/necromorph/corruption/growth.dm
Ketrai: This is fine.

code/modules/necromorph/corruption/growth.dm
Ketrai: Nanako, the main issue is that roots support UP TO 40 RANGE OF TILES. 
The way this looks in practice? You place a root, everywhere within 40 tiles starts to spread at an incredible fast rate. Humans delete a root? it keeps going, because there's likely a root somewhere within 40 tiles that keeps the corruption there. This is absolutely needed. It doesn't even spread well in maint, because it might be expanding off the other end of the ship, because again, 40 tile range. it's instead being used as the only root, because it just spreads faster with the slight negative of requiring a few more nodes than branches.

ghost:
It's an easy fix to reverse if need be. I'll be merging this for now but if it needs adjustment and Necros are getting rolled @Ketrai, I want us to adjust immediately. Approved provisionally.

PR #1874, opened 2021-10-13
Fixes necros spawning outside in a tram spawn.

Ketrai:
Necros were spawning in space here and often breached the ship because of it.

### DIFF
--- a/maps/DeadSpace/M.02 Ishimura (T).dmm
+++ b/maps/DeadSpace/M.02 Ishimura (T).dmm
@@ -388,10 +388,15 @@
 /turf/simulated/floor/plating,
 /area/ishimura/tramdeck/tram/misc/disusedbotanics)
 "bd" = (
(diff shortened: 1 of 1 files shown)


NanakoAC:
This seems like a code problem rather than mapping

can you show what exactly the issue is, and what you've done to fix it?

Ketrai:
> This seems like a code problem rather than mapping
> 
> can you show what exactly the issue is, and what you've done to fix it?

It looks like a mapping error coz there's some non-space area outside the ship. Where they were spawning. Otherwise necros don't spawn outside.

Snypehunter007:
I can fix this on my current tram deck branch and merge it when I finish that if you can give me a picture of the problem, but I'm waiting for someone to fix double doors before I can proceed.

NanakoAC:
closing since snype says he'll fix it

PR #1878, opened 2021-10-15
SSresearch and some minor optimization [DNM]

DTraitor:
### Purpose
SSresearch now actually handles designs, technologies, trees, research design asset and other stuff. It takes ~1 second on my PC to initialize subsystem (just for reference, it takes 44-48 seconds to init SSatoms). This time can be lowered by adding unique ids to all autolathe designs and specifying time to print thing (right now it's calculated during `AssembleDesignInfo()`.
After this PR you should avoid spawning datum/design directly or if it isn't in the list you should spawn it (**MAKE SURE IT HAS UNIQUE ID AND BUILD PATH**) and call `SSresearch.register_research_design(datum/design/D)`.
The only place that actually references to designs instead of storing ids is SSdatabase (I didn't want to touch it since I don't have database to test store, rivet gun design seems to work) but those designs are same datums as SSresearch stores.

This PR reduces ammount of datums **A LOT** (in current dev branch it is over 300 datums in SSresearch, and over 400 datums per every `/datum/research` (every server and rdconsole has it, we have 1 console and 1 server on Ishimura)). Now it is's over 400 datums in SSresearch and few lists with ids in `/datum/research`.

Tell me if I should add comments to SSresearch.


code/controllers/subsystems/characters/database_persistence.dm
```diff
@@ -185,7 +185,7 @@ SUBSYSTEM_DEF(database)
 	unknown_designs = list()
 
 	//This gets a list of every design that exists
-	var/list/designs = SSresearch.design_ids.Copy()
+	var/list/designs = SSresearch.design_by_id.Copy()
```
NanakoAC: this seems an unnecessary change
at least make it "designs_by_id"

DTraitor: "designs_by_id" makes more sense. Var was renamed because of three reasons: "design_ids" doesn't describe actuall purpose, "design_by_id" was var that existed in `/datum/research`, I had to check if all cases where it is required use it in right way.

PR #1883, opened 2021-10-16
Hydrazine torch fires in vacuum again

### LINKED ISSUE #1856: [Bug] Hydrazine torch can not fire Hydrazine in a vacuum

DTraitor:
### Purpose
Fixes: https://github.com/DS-13-Dev-Team/DS13/issues/1856

### DIFF
--- a/code/modules/projectiles/guns/ds13/flamethrower.dm
+++ b/code/modules/projectiles/guns/ds13/flamethrower.dm
@@ -11,7 +11,7 @@
 	//This gun has no safety switch
 	safety_state = FALSE
 
-	var/obj/item/weapon/reagent_containers/glass/fuel_tank/tank
+	var/obj/item/weapon/reagent_containers/glass/fuel_tank/tank = /obj/item/weapon/reagent_containers/glass/fuel_tank
 
 	//25 Seconds continuous firing from a 400u fueltank
 	var/fuel_per_second = 16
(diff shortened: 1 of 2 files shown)


Ketrai:
Make sure it's only for hydrazine fuel though. Not regular fuel.

code/modules/projectiles/guns/ds13/flamethrower.dm
```diff
@@ -431,7 +431,6 @@
 */
 /obj/item/weapon/reagent_containers/glass/fuel_tank/proc/check_fuel_type_and_contamination()
 	fueltype = /datum/reagent/fuel
-	vacuum_burn = FALSE
```
NanakoAC, line 434: i don't see how this would help. The same logic still applies doesn't it?

code/modules/projectiles/guns/ds13/flamethrower.dm
```diff
@@ -468,6 +468,7 @@
 
 /obj/item/weapon/reagent_containers/glass/fuel_tank/hydrazine
 	name = "fuel tank (hydrazine)"
+	vacuum_burn = TRUE
```
NanakoAC: this shouldn't be necessary as the fuel updating proc handles it. If there's a situation where its not being called, find that and put in a call

NanakoAC (changes requested):
After discussion in dev closet, the recommended solution here is as follows:

Either Add a new gun proc to create the initial ammo, or make the process of creating initial ammo call something (ammo inserted?) that is also called when a magazine/tank/bullet/etc is inserted normally

Override that proc to make the hydrazine torch update its fuel when it recieves a new tank

NanakoAC (approved):
lets just get this merged

PR #1892, opened 2021-10-17
Resolve #1820:

### LINKED ISSUE #1820: Vector Invalid Dir
[15:13:33] Runtime in CommonVectors.dm, line 50: Invalid direction.
proc name: FromDir (/Vector2/proc/FromDir)
src: /Vector2 (/Vector2)
call stack:
/Vector2 (/Vector2): FromDir(16)
Jason Augment (/mob/living/carbon/human/necromorph/twitcher): get view centre()
Jason Augment (/mob/living/carbon/human/necromorph/twitcher): turfs in view(null)
Jason Augment (/mob/living/carbon/human/necromorph/twitcher): get visualnet tiles(/datum/visualnet/necrovision (/datum/visualnet/necrovision))
/datum/chunk/necrovision (/datum/chunk/necrovision): get datum visible turfs(Jason Augment (/mob/living/carbon/human/necromorph/twitcher))
/datum/chunk/necrovision (/datum/chunk/necrovision): acquire visible turfs(/list (/list))
/datum/chunk/necrovision (/datum/chunk/necrovision): update(0)
Chunks (/datum/controller/subsystem/chunk): fire(0)
Chunks (/datum/controller/subsystem/chunk): ignite(0)
Master (/datum/controller/master): RunQueue()
Master (/datum/controller/master): Loop()
Master (/datum/controller/master): StartProcessing(0)

Kmc2000:
-get_view_centre() could be passed an invalid dir.
-turfs_in_view utilised a static singleton mob for its checking beyond its intended purpose (lighting code) thereby opening up a slew of concurrency issues. This has been resolved by simply running a view() check on the caller, as there were no modifications necessary for its see_invisible flags (it was previously always called with 0).

If the use of GLOB.dview_mob is absolutely necessary for a reason I don't understand please do correct me.

### DIFF
--- a/code/_helpers/detection.dm
+++ b/code/_helpers/detection.dm
@@ -58,31 +58,35 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	return view(range, GLOB.dview_mob)
 
 //Returns a list of all turfs this mob can see, accounting for view radius and offset
-/atom/proc/turfs_in_view(var/check_range = world.view)
-	var/list/things = list()
-	FOR_DVIEW(var/turf/T, check_range, get_turf(src), 0)
-		things += T
-	END_FOR_DVIEW
-	return things
+/atom/proc/turfs_in_view(var/check_range = world.view, var/atom/origin=get_turf(src))
+	RETURN_TYPE(/list)
+	var/list/ret = list()
+	for(var/T as () in view(check_range, origin))
+		if(!isturf(T))
+			continue
+		ret += T
+	return ret
 
-/mob/turfs_in_view(var/check_range = null)
-	var/range = (check_range ? check_range : view_range)
-	if (!isnum(range) || range < 0)
-		range = world.view
-	else
-		range = Ceiling(range)	//Just incase a noninteger value was passed
-	var/origin
-	if (!view_offset)
-		origin = get_turf(src)
-	else
-		origin = get_view_centre()
+/**
+<summary>Gets all the turfs in view of this mob. Ignores invisibility as it was never used.</summary>
+<returns>A list of all the visible turfs for this mob, as /turf/.</returns>
+*/
 
-	var/list/things = list()
-	FOR_DVIEW(var/turf/T, range, origin, 0)
-		things += T
-	END_FOR_DVIEW
-	return things
-
+/mob/turfs_in_view(var/check_range = view_range)
+	//Handle unsafe int32 values.
+	check_range = (!isnum(check_range) || check_range < 0) ? world.view : Ceiling(check_range)
+	var/turf/_origin = (view_offset) ? null : get_turf(src)
+	//Try get the offset from the vector. This can crash if there's an invalid dir of "16". If this fails, we just go with pure loc. If _origin is null, we have a view offset to account for.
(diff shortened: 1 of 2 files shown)


code/_helpers/detection.dm
```diff
@@ -58,31 +58,35 @@ GLOBAL_DATUM_INIT(dview_mob, /mob/dview, new)
 	return view(range, GLOB.dview_mob)
 
 //Returns a list of all turfs this mob can see, accounting for view radius and offset
-/atom/proc/turfs_in_view(var/check_range = world.view)
```
NanakoAC, line 61: The intended purpose of dview is to see in the dark, ignoring byond-native darkness that obscures things. Because it is an unreliable and largely obsolete interference we don't use

Also a secondary consideration, the possibility that the atom is inside something, we generally want this proc to return a view from the toplevel.

Lastly though, i think the conclusion of concurrency issues is questionable. DVIEW operates in a single frame as far as i'm aware, i don't see any possibility of interference even if a million things are all trying to use it

Kmc2000: The initial proc just set sight_flags to 0 every time rather than utilising the sight flags. A view from the top-level is a valid argument, I can account for this.
RE: Concurrency. I still think it's a possible explanation for this problem. That dview mob was intended for lighting (citation needed), not to be accessed by a load of mobs simultaneously. The START_DVIEW macro makes no check / consideration for whether the dview mob is actually available, and it's entirely possible that two concurrent calls to it could cause this weirdness with dir. It may well operate in a single frame, but what about the callers? under load, is it not feasible that two mobs could have their calls to it queued up and executed at almost exactly the same time.

NanakoAC: The dview mob was intended for getting view information WITHOUT lighting. The D stands for dark

I'm pretty sure byond is singlethreaded, two concurrent calls is not actually a thing that can happen. Unless sleeps are involved, i don't see any way that the dview mob could fail to finish one operation before another starts.

lib/vector2/CommonVectors.dm
```diff
@@ -27,6 +29,9 @@ GLOBAL_REAL(Vector2, /Vector2) = new
 	Northwest = new(-sqrt(1/2), sqrt(1/2))
 	Southeast = new(sqrt(1/2), -sqrt(1/2))
 	Southwest = new(-sqrt(1/2), -sqrt(1/2))
+	//DS13 Change - Multiz support.
+	Up = new(1, 1)
```
Kmc2000, line 33: Not 100% on these values, could use some insight here. They're mostly just placeholder.

NanakoAC: nah it doesn't make sense to have vectors for up and down at all, since they represent movement in the third dimension

NanakoAC (changes requested):
Dtraitor found out where its inserting vertical directions for view checking

Vertical movement should just never get to the point where it tries to fetch a vector dir, cut it off at that point

NanakoAC (changes requested):
actually on second thought, your solution of adding in vectors for up and down is fine, but make those vectors be 0,0

and make sure the source which could possibly pass in up/down can handle that return value

PR #1894, opened 2021-10-18
Corruption Tweaks

NanakoAC:
changes:
  - balance: "Reduced the energy cost of snare and audio abilities for signals."
  - balance: "Slightly reduced the health and armor of the Harvester node."
  - balance: "Corruption now spreads more slowly."
  - balance: "Reduced the minimum placement buffer of snare to 2 tiles."
  - bugfix: "Corruption will no longer grow over openspace tiles"

### DIFF
--- a/code/controllers/configuration/entries/general.dm
+++ b/code/controllers/configuration/entries/general.dm
@@ -64,6 +64,7 @@ Basics, the most important.
 /datum/config_entry/flag/allow_vote_mode    //allow votes to change mode
 
 /datum/config_entry/flag/allow_admin_jump   //allows admin jumping
+	config_entry_value = TRUE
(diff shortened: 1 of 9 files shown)


code/controllers/configuration/entries/general.dm
```diff
@@ -64,6 +64,7 @@ Basics, the most important.
 /datum/config_entry/flag/allow_vote_mode    //allow votes to change mode
 
 /datum/config_entry/flag/allow_admin_jump   //allows admin jumping
+	config_entry_value = TRUE
```
NanakoAC, line 67: setting this default true because it was getting annoying, admins shouldn't be prevented from jumping

PR #1910, opened 2021-10-20
banditofdoom's special rig

NanakoAC:
Another custom item enters


NanakoAC:
quickmerging this to get it in for his birthday

PR #1920, opened 2021-10-24
Atom_colours, filter_data and optimized rendering.

DeadLineSS13:
## About the PR.
This PR added filtered lists atom_colours, filter_data and strongly and optimized rendering plane_master.

Now better use add_filter when filters += something. (You can ovverride procs adding and updating)
And better use add_atom_colour when color += something. (You can ovverride procs adding and updating)
obj/screen now atom/movable/screen.

## Changelog
🆑 DeadLineSS13
refactor: Refactored code with Filters, Colours and Rendering on game.
/🆑


code/_helpers/filters.dm
```diff
@@ -0,0 +1,319 @@
...
+	/* Not supported because making a proper matrix editor on the frontend would be a huge dick pain.
+		Uncomment if you ever implement it
+	"color" = list(
+		"defaults" = list(
+			"color" = matrix(),
+			"space" = FILTER_COLOR_RGB
+		)
+	),
+	*/
+	"displace" = list(
```
NanakoAC, line 34: doesnt this data already exist in the hidden sddef file?

code/__defines/misc.dm
```diff
@@ -270,3 +270,10 @@
 #define STATION_ALERT_BLUE	"code blue"
 #define STATION_ALERT_RED	"code red"
 #define STATION_ALERT_DELTA	"code delta"
+
+//different types of atom colorations
+#define ADMIN_COLOUR_PRIORITY 1 //only used by rare effects like greentext coloring mobs and when admins varedit color
+#define TEMPORARY_COLOUR_PRIORITY 2 //e.g. purple effect of the revenant on a mob, black effect when mob electrocuted
+#define WASHABLE_COLOUR_PRIORITY 3 //color splashed onto an atom (e.g. paint on turf)
+#define FIXED_COLOUR_PRIORITY 4 //color inherent to the atom (e.g. blob color)
+#define COLOUR_PRIORITY_AMOUNT 4 //how many priority levels there are.
```
NanakoAC, line 279: i really don't like this. Extensions with statmods would be a far better way of handling the few edge cases of icons that get colored more than once

code/_onclick/hud/ability_screen_objects.dm
```diff
@@ -32,21 +32,21 @@
...
 
-/obj/screen/movable/ability_master/Click()
+/atom/movable/screen/movable/ability_master/Click()
 	if(!ability_objects.len) // If we're empty for some reason.
 		return
 
 	toggle_open()
 
-/obj/screen/movable/ability_master/proc/toggle_open(var/forced_state = 0)
+/atom/movable/screen/movable/ability_master/proc/toggle_open(var/forced_state = 0)
```
NanakoAC, line 47: why this mass retyping?  Atom movable is a shared parent of objects and mobs, is there some kind of mob behaviour you want to add to them?

code/game/atoms.dm
```diff
@@ -33,11 +33,20 @@
 	var/default_alpha = 255
 	var/default_scale = 1
 
+	/**
+	 * used to store the different colors on an atom
+	 *
+	 * its inherent color, the colored paint applied on it, special color effect etc...
+	 */
+	var/list/atom_colours
```
NanakoAC, line 41: this snowflake system seems inferior to using extensions, i really don't like it

code/_helpers/filters.dm
```diff
@@ -0,0 +1,319 @@
...
+			"y" = 0
+		)
+	),
+	"outline" = list(
+		"defaults" = list(
+			"size" = 0,
+			"color" = COLOR_BLACK,
+			"flags" = NONE
+		),
+		"flags" = list(
```
Kmc2000, line 81: Does this really need to be a list? This looks like it can be made into a bitflag

DeadLineSS13: > Does this really need to be a list? This looks like it can be made into a bitflag

I agree with you, but at the moment for Byond it will be perceived as list strings. You can check this yourself by decoding the filter "outline".

NanakoAC (approved):

PR #1923, opened 2021-10-25
Speed and Health

NanakoAC:
Changes aimed at shifting the metagame based on discussion with staff yesterday

Two broad categories: Nerfing the speed of powergamers, and increasing TTK across the board

changes:
  - experiment: "Reduced human top speed. The intent here is to nerf the most extreme examples of zoomy humans without significantly affecting the majority of people. This is broken up into several changes:"
  - balance: "Reduced the speed bonus of advanced femoral exoskeleton, Basic femoral unaffected"
  - balance: "Slightly Increased stamina costs of sprinting, and slightly reduced stamina regen"
  - balance: "Athletics skill is now broken into two skills, Hauling and Athletics. The former affects speed penalties from encumbrance, throwing, and pulling, the latter affects stamina costs."
  - balance: "Slightly increased corruption slowdown effect on humans."
  - balance: "Slightly decreased sprint speed, and slightly increased walking speed."
  - experiment: "Increased Time To Kill (TTK) by 10% across the board. This is broken up into several changes:"
  - balance: "Increased human limb health by 10%"
  - balance: "Increased human pain-shock threshold by 10%. from 50 to 55 This is the quantity of pain that can be suffered without any real bad effects."
  - tweak: "Increased health of all necromorphs by 10%"

### DIFF
--- a/code/__defines/movement.dm
+++ b/code/__defines/movement.dm
@@ -25,10 +25,10 @@
 
 
 
-#define RUN_DELAY 1.5
-#define WALK_DELAY 3
+#define RUN_DELAY 1.6
+#define WALK_DELAY 2.95
 #define STALK_DELAY 6
-#define MINIMUM_SPRINT_COST 0.8
-#define SKILL_SPRINT_COST_RANGE 0.8
-#define MINIMUM_STAMINA_RECOVERY 2
-#define MAXIMUM_STAMINA_RECOVERY 6
+#define MINIMUM_SPRINT_COST 0.85	//The part of sprint cost that everyone gets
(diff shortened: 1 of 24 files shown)


Ketrai:
A bigger corruption slowdown is more annoying than anything. Humans get punished enough for going on corruption. Necros also get sped up at the same time.

I don't think necromorphs need more health? I'd rather nerf problem weapons or buff necromorphs in other ways. Just because a flamethrower destroys necromorphs in seconds, doesn't mean we should make it harder for plasma cutters to kill.

Snypehunter007:
@NanakoAC tweak no longer works due to Deadline's Change Log changes.

PR #1925, opened 2021-10-25
SStimer runtimes fix

DTraitor:
### Purpose
Fixes a lot of SStimer runtimes. Most likely they were caused by changing FPS during world initialization instead of setting it in the code. This shouldn't affect anything, I haven't seen any timer runtimes playing this branch but it's possible that they will exist after this PR since runtimes are randomly triggered and I was just unlucky in reproducing any of them.

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -11,7 +11,7 @@
 	cache_lifespan = 7
 	hub = "Exadv1.spacestation13"
 	icon_size = WORLD_ICON_SIZE
-	fps = 20
+	fps = 50
 #ifdef GC_FAILURE_HARD_LOOKUP
 	loop_checks = FALSE
 #endif


NanakoAC (approved):

code/world.dm
```diff
@@ -11,7 +11,7 @@
 	cache_lifespan = 7
 	hub = "Exadv1.spacestation13"
 	icon_size = WORLD_ICON_SIZE
-	fps = 20
+	fps = 50
```
Kmc2000, line 14: Are you sure about this? Is this not going to force the ticker to run way faster now?
Also, do you have any reason for this specific number, or is it just taken from somewhere like /tg/

DTraitor: We use 50 fps already but it wasn't hard-defined

PR #1941, opened 2021-11-04
SSTicker, SSVote with first fixes.

DeadLineSS13:


NanakoAC (approved):
this is already live so no reason not to approve

PR #1952, opened 2021-11-12
Smetler fixes and tweaks

### LINKED ISSUE #1951: Bug: Ore smelter stops processing at some point
#### Description of issue:
If there is a lot of ore (over 1.5k) in smelter input tile, it can stop processing.
This is just a guess but it may be connected with how locate() works.

Smelter shouldn't stop processing if it has a lot of ore in input tile.

#### Steps to reproduce:
Mine a lot of ore and try to smelt it.

DTraitor:
### Purpose
Fixes: #1951
Better scanning modules will speed up ore gathering.

### DIFF
--- a/maps/DeadSpace/M.03 Ishimura (L).dmm
+++ b/maps/DeadSpace/M.03 Ishimura (L).dmm
@@ -26800,13 +26800,13 @@
 /turf/simulated/floor/dank/golf_brown,
 /area/ishimura/lower/mining/processing_room)
 "bdy" = (
-/obj/machinery/input,
 /obj/effect/floor_decal/industrial/loading{
 	dir = 4
 	},
 /obj/machinery/light{
 	dir = 1
 	},
+/obj/machinery/input/ore,
 /turf/simulated/floor/dank/rectangles,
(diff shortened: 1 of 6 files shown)


maps/DeadSpace/M.03 Ishimura (L).dmm
```diff
@@ -26800,13 +26800,13 @@
...
 /area/ishimura/lower/mining/processing_room)
 "bdy" = (
-/obj/machinery/input,
 /obj/effect/floor_decal/industrial/loading{
 	dir = 4
 	},
 /obj/machinery/light{
 	dir = 1
 	},
+/obj/machinery/input/ore,
```
NanakoAC, line 26809: why does a new machine need to be put in?

Not that i'd mind, but mapping changes can cause conflicts with other ongoing changes to the same map

DTraitor: `/ore` subtype calls `wake_up()` in `Uncrossed()` too.

NanakoAC (approved):
snype can overwrite the mapping change and reimplement it, he's been informed

PR #1965, opened 2021-11-23
Replaces a lot of istype() with proper defines

DTraitor:


code/modules/hydroponics/grown_inedible.dm
```diff
@@ -40,12 +40,12 @@
...
 	w_class = ITEM_SIZE_SMALL
 	throwforce = 0
-	
+
 	throw_range = 20
 
 /obj/item/weapon/corncob/attackby(obj/item/weapon/W as obj, mob/user as mob)
 	..()
-	if(istype(W, /obj/item/weapon/tool/saw/circular) || istype(W, /obj/item/weapon/material/hatchet) || istype(W, /obj/item/weapon/material/kitchen/utensil/knife) || istype(W, /obj/item/weapon/material/knife) || istype(W, /obj/item/weapon/material/knife/ritual))
+	if(isSaw(W) || istype(W, /obj/item/weapon/material/hatchet) || istype(W, /obj/item/weapon/material/kitchen/utensil/knife) || istype(W, /obj/item/weapon/material/knife) || istype(W, /obj/item/weapon/material/knife/ritual))
```
NanakoAC: this section should require any blade rather than all these checks, 
and just check for the cutting tool property on W

NanakoAC (changes requested):
just the pipe blades thing

PR #1983, opened 2022-01-10
Icons clean up and proper tool icons

DTraitor:
#1955


code/datums/sound_player.dm
```diff
@@ -215,7 +215,7 @@ datum/sound_token/proc/Mute()
 	status = new_status
 	PrivUpdateListeners()
 
-/datum/sound_token/proc/PrivAddListener(var/atom/listener)
+/datum/sound_token/proc/PrivAddListener(var/mob/listener)
```
NanakoAC, line 218: I don't think this change makes sense
isn't this system designed to accomodate simulating objects recieving sounds? 
clients could also be passed in

code/datums/sound_player.dm
```diff
@@ -224,6 +224,9 @@ datum/sound_token/proc/Mute()
 	if(listener in listeners)
 		return
 
+	if(!istype(listener, /mob))
```
NanakoAC, line 227: this safety check seems unnecessary, and it does overlap with the virtualmob check above, consider reordering things here

NanakoAC (commented):
a little confusion about the sound tokens, also should check with snype regarding the little mapping change

PR #1987, opened 2022-01-13
Sound QoL

ghost:
-Added or edited sounds for interacting with many objects. Smartfridges, fire alarms, portable generators, portable air pumps, portable scrubbers, disposal chutes, atmos canisters, atmos equipment, pipe dispensers, emitters/gyrotrons... etc etc

The intent is to remove 'silent' objects that have absolutely zero feedback

-Fixed footstep sounds for the new themed flooring.

-Growth tank bubbles had a volume increase, it's still subtle but you couldn't even hear it before on some headphones.

-Evidence lockers are now their own independent object rather than brig lockers with prisoner clothing in them for no reason.


NanakoAC (approved):
amazing work, just the kind of thing we need

PR #2035, opened 2022-04-20
[DNM] Test Merge Only

DTraitor:
Gonna track every case where human is deleted to figure out what causes random deletions.

P.S. Yes that's kinda dumb way to do that but we need to know what calls qdel()


DTraitor:
We found the issue

PR #2042, opened 2022-04-24
Gives Aegis breathable atmos

Ketrai:
As requested by DTraitor, let there be breath.

Airlock is still there but works more like a door now. It's nice flavor either way.


DTraitor (approved):
Looks good to me

PR #2045, opened 2022-04-25
Twitcher Icons

Ketrai:
New twitcher sprite!


Ketrai:
This also fixes necro sprites getting mangled visually when you delimb anything. Except for slashers because it would break their clothes. But at least this is a partial fix for now.

PR #2056, opened 2022-05-01
Buffs the undertram spawn for the marker

Ketrai:
Adds a couple more valid machines for the marker to leech off to compensate for the much slower corruption spread. Adds a couple edibles to absorb, but it's only about 10KG at best, more flavor than anything impactful. 

Removes one of the marker spawns down there as well. I don't know why I put two down there, that's a bit silly.

- Also removes a large amount of metal foam. They were double stacked in a lot of places. (How, why lion?) And it's arthritis inducing to try and click all of them away to reveal sources. 

(Click to enlarge for a less blurry preview)

image

(Oh and the human shape is a monki)


Ketrai:
image
With less foam

PR #2079, opened 2022-05-03
Reworks enhanced leaper

Ketrai:
- balance: "Made enhanced leaper more agile."
  - balance: "More than halves enhanced leaper damage."
  - balance: "Lowered leaper damage slightly."
  - balance: "Leaper is now unable to gallop continuously, there's downtime."
  
  Enhanced leapers are now quick, but their damage is far less than that of a regular leaper. This should make them rely more on cooperation with other necros, especially regular leapers.
  
  Regular leapers now can't gallop basically without cooldown. They've got an actual cooldown to their tail strike as well so it's unable to be spammed anymore. This should make them meme doors less quickly, and give survivors more of a chance to recover from being leapt on at close range.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/necromorph/leaper.dm
+++ b/code/modules/mob/living/carbon/human/species/necromorph/leaper.dm
@@ -34,7 +34,7 @@
 
 	evasion = 20	//Harder to hit than usual
 	var/stun = 2 //stun duration
-	var/leap_damage = 3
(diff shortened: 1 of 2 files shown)


Ketrai:
Closing this for now, merge #2080 instead.

Leapers will probably need more tweaks but I'm giving it a couple days for people to give feedback.

PR #2081, opened 2022-05-04
Stops age from affecting skills.

Ketrai:
I'm not sure why bay ever added this given this is purely a ticket to powergame. Even on a HRP server you're going to see an above-average amount of old farts purely because they get more skill points.

If jobs need more skill points we can set this for individual jobs instead.


DTraitor:
F

PR #2087, opened 2022-05-05
It twitches...

Ketrai:
Adds the oracle twitcher.

Mostly the same as a regular twitcher, except it blinks to dodge attacks much more often, and it's quite a bit faster. (Twitchers are already strong enough so I am very cautious about making this necro too strong.) The slash damage is slightly slower but there's little cooldown between attacks, allowing it to finish off targets quick. This also means the blink and step strike damage is less, since they're based off the damage dealt by twitcher's main weapons.


DTraitor (approved):
Looks good to me (if we can say anything looks good in our codebase at all)

Issue #2106, opened 2022-05-23
FEATURE ADJ: Stores System and Detaching from Persistent Economy


ghost:
#### Their current behavior:

Right now, the way Stores work is that you acquire credits via your paycheck or by finding other credits/peng, and redeeming them at Stores for items. Credits can persist between rounds using a database. They work as a one-size-fits-all vendor, and you can even hot-swap RIGs at them. Credits are tied to a person's RIG.

#### Their intended behavior:

Stores should be fully detached from the persistent economy, keeping purchases and credits earned only applicable to the round they're earned in. Credits should be earned via the initial paycheck, AND through killing necromorphs, who drop varying degrees of credits. 

Stores should not have a giant swath of items that you can otherwise get at free vendors elsewhere.
Stores should be the best possible access point to purchasing actual firearms, RIGs, ammunition, medicine and other high-end rewards.
Stores should still be able to hot-swap RIGs.
Credits should probably not be tied to RIGs anymore, and should instead exist as their own stackable item or be tied to something else entirely, if possible. 
** Separate but also important, the mining system for redeeming credits should be detached from the main economy, so that miners can't flood the game with easily acquired credits. Their system of rewards should be unique to them and through a different vendor.

- [x] Take the existing Stores and turn into an independent vendor severed from the persistent economy system.
- [x] Erase or comment out current persistent economics code (ideally without breaking everything).
- [ ] Adjust the available items list in Stores to only include Weapons, RIGs, Modules, Medicine, Ammunition and Rare Items.
- [ ] Adjust the prices of those items by 30-50%.
- [ ] Have necromorphs drop a certain number of credits, adjust drop amount for balancing. 
- [ ] Detach credits from RIGs as their primary means of being stored -- ensure credits are stackable item that can be stored in a backpack or taken out of an ATM. (think caps from Fallout 13)
- [x] Ensure hot-swap mechanics from Stores for changing RIGs is still usable.

Viz-Titan:
There are also set map drop spawns of credits to consider. Some places, like Aft Janitorial Closet room in maints above Engineering Breakroom on Upper, spawn a big load of credits. Around 20k or so worth at that spot. 
It is nothing vital but just something to note for much later. Wouldn't be too hard to fix/balance their spawns. Could just get rid of those spawns and replace with low values or tweak the credit values on them to be lower. Or get rid of credit spawns entirely.
[...]

PR #2119, opened 2022-06-18
Detaching from Persistent Economy

SirRichardFrancis:
## About The Pull Request

Does what #2106 says, with an exception of adjusting available item lists and prices, same with mining vendor, it only contains couple of power nodes right now.
Haven't touched the balancing part, since I'm barely familiar with the build and it's IC economy.


code/__defines/research.dm
```diff
@@ -12,10 +12,11 @@
...
-#define STORE		32	//Purchasing in store
\ No newline at end of file
+#define NONE				0x1
+#define IMPRINTER			0x2		// For circuits. Uses glass/chemicals.
+#define PROTOLATHE			0x4		// New stuff. Uses glass/metal/chemicals
+#define MECHFAB				0x8		// Mechfab
+#define CHASSIS				0x10	// For protolathe, but differently
+#define AUTOLATHE			0x20	// For autolathe
+#define STORE_SCHEMATICS	0x40	// Bound to schematic, unlockable in store with said chematic
+#define STORE_ROUNDSTART	0x80	// Always available in store
```
DTraitor: use `(1<<X)` instead

code/controllers/subsystems/ticker.dm
```diff
@@ -188,6 +187,8 @@ SUBSYSTEM_DEF(ticker)
 	log_world("Game start took [(world.timeofday - init_start) / 10]s")
 	round_start_time = world.time
 
+	GLOB.using_map.setup_economy()
```
DTraitor: Remove this one, I've added `setup_economy()` to `/datum/map/proc/setup_map()` some time ago

code/modules/economy/credits.dm
```diff
@@ -30,7 +22,7 @@
 /obj/item/weapon/spacecash/ewallet/update_icon()
 	icon_state = "grey"
 	switch(worth)
-		if (1 to 500)
+		if (0 to 500)
```
DTraitor: It should be `icon_state = "grey"` if there are no credits

code/modules/economy/store/store_credits.dm
```diff
@@ -23,49 +31,52 @@
...
+		var/pay_amount = min(occupant.wearing_rig.get_account_balance(), left_to_pay)
 		var/obj/item/weapon/rig/R = occupant.wearing_rig
 		R.charge_to_rig_account(src, "Store Purchase", machine_id, -pay_amount)
-		cost -= pay_amount
+		left_to_pay -= pay_amount
 
+	if(left_to_pay)
+		for(var/i in deposit_box.contents)
+			if(istype(i, /obj/item/weapon/spacecash))
+				var/obj/item/weapon/spacecash/S = i
```
DTraitor, line 56: Use `for(var/obj/item/weapon/spacecash/S in deposit_box.contents)` instead, for loop + istype is faster only for short types (according to my tests)

code/modules/economy/store/store_credits.dm
```diff
@@ -23,49 +31,52 @@
...
+					S.worth -= left_to_pay
+					S.update_icon()
+					deposit_box.update_ui_data()
+					return TRUE
 
-	if (cost <= 0)
-		return TRUE
+	var/datum/money_account/A = occupant?.get_account()
+	if(A)
+		A.money -= left_to_pay
```
DTraitor, line 69: There are no safety checks here meaning our account balance can become lower than 0

code/modules/economy/store/store_credits.dm
```diff
@@ -115,8 +126,10 @@
 /obj/machinery/store/proc/buy_and_transfer()
 	var/list/things = list() + buy_current()
 	var/obj/item/weapon/rig/R
-	for (var/obj/item/I in things)
-		if (isrig(I))
+	if(occupant.back && !occupant.wearing_rig)
+		store_or_drop(occupant.back)
+	for(var/obj/item/I in things)
+		if(isrig(I))
 			R = I
```
DTraitor: Just use `var/obj/item/weapon/rig/R = locate(/obj/item/weapon/rig) in things` instead of for loop

DTraitor (changes requested):
Just a few tweaks

code/modules/economy/store/store_credits.dm
SirRichardFrancis: We get here only if `occupant_can_afford()` returned `TRUE`, meaning user have enough money, but no money sources other than this one left, hence check here is redundant, as there is no cases for 'not enough money' at this point.

code/modules/economy/store/store_credits.dm
SirRichardFrancis: There is `store_or_drop(I)` called for every item in `things`, although I'm not sure if there is a situation when that list contain more than one object.

DTraitor: In that case use `locate()` then call `start_transfer(R)` and remove it from `things`, after that we can loop through all items in `things` without checking if it's a RIG.

DTraitor (approved):
Looks good now

PR #2143, opened 2022-08-11
QOL/BUGFIX/BALANCE changes!

karlodayo:
Fixes necromorph chat color, toggable clothing, and surgeon instakill.


code/modules/clothing/suits/labcoat.dm
```diff
@@ -25,7 +25,7 @@
 	name = "chief medical officer's labcoat"
 	desc = "Bluer than the standard model."
 	icon_state = "labcoat_cmo_open"
-	//item_state = "labcoat_cmo"
+	// item_state
```
DTraitor: Just remove this comment and it's ready to merge

DTraitor (changes requested):

PR #2157, opened 2022-08-27
Reworking system FREE SKILL POINTS. Now we have Background system.

DeadLineSS13:
## ABOUT THE PR
Refactored moment, what player can use free skill points. Now players need choose background history character for them builds.

## CHANGELOG
🆑 
refactor: Now we don't have Free Skill Points. WELCOME NEW BACKGROUND SYSTEM!!!!!
/🆑


Vizoplez:
This is literally just porting a sojourn-esque skill system that will encourage people to only pick military, as it is the only stat that matters. CQC and Weapons experience.

This... fixes nothing and adds nothing to the admittedly already-broken RP loop.

Why?

DeadLineSS13:
> This is literally just porting a sojourn-esque skill system that will encourage people to only pick military, as it is the only stat that matters. CQC and Weapons experience.
> 
> This... fixes nothing and adds nothing to the admittedly already-broken RP loop.
> 
> Why?

People can't be good at CQC and Weaponary Experiense at the same time. Since for CQC you still need to have other skills, but for Weaponary Experiense this is not necessary. We also do not forget that if they give up on medicine, then who will treat? Many factors affect the gameplay of the game. In this PR, the theme of free Skill Points will simply be killed, so that a medic, for example, would not be a master in combat, in weapons. I agree that PR is not balanced, but it cannot be balanced at the level of everyday play, therefore + - stats are thrown.

DTraitor (approved):

Vizoplez:
> > This is literally just porting a sojourn-esque skill system that will encourage people to only pick military, as it is the only stat that matters. CQC and Weapons experience.
> > This... fixes nothing and adds nothing to the admittedly already-broken RP loop.
> > Why?
> 
> People can't be good at CQC and Weaponary Experiense at the same time. Since for CQC you still need to have other skills, but for Weaponary Experiense this is not necessary. We also do not forget that if they give up on medicine, then who will treat? Many factors affect the gameplay of the game. In this PR, the theme of free Skill Points will simply be killed, so that a medic, for example, would not be a master in combat, in weapons. I agree that PR is not balanced, but it cannot be balanced at the level of everyday play, therefore + - stats are thrown.

(Foreshadowing)

PR #2171, opened 2022-09-29
Nerfs tool mods

Ketrai:
Makes tool mods slightly less insane. These stack multiplicatively for some reason.
Also makes the side-effects less bad to compensate.

No more 50-60 damage saws.


DTraitor (approved):
Tbh you could just remove craft recipe that allows to stack so many mods

PR #2196, opened 2022-10-15
Reduces the volume of salt in salt sacks

githubuser4141:
What this PR Does & Why:

Simply put, it reduces the amount of salt found in the large salt containers in spawn lockers to 50 units, from 500, and adds a warning note.

Why? 'Cause being permanently round-removed from excessive salt consumption isn't fun, and, there's really no reason for such a huge amount of salt to exist round start. It makes ordering salt from Supply pointless as there's enough salt in each bag to last the entire shift even if you're an active cook.

EDIT: Crew & Others can munch on spicy food without fear of stomach ulcers now! :) (No idea if that's a thing in the code or not.)

### DIFF
--- a/code/modules/reagents/reagent_containers/food/condiment.dm
+++ b/code/modules/reagents/reagent_containers/food/condiment.dm
@@ -96,7 +96,8 @@
 				center_of_mass = "x=16;y=6"
 			if(/datum/reagent/nutriment/soysauce)
 				name = "Soy Sauce"
-				desc = "A salty soy-based flavoring."
+				desc = "A 
+				y soy-based flavoring."
 				icon_state = "soysauce"
 				center_of_mass = "x=16;y=6"
 			if(/datum/reagent/frostoil)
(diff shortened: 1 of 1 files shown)


code/modules/reagents/reagent_containers/food/condiment.dm
```diff
@@ -96,7 +96,8 @@
 				center_of_mass = "x=16;y=6"
 			if(/datum/reagent/nutriment/soysauce)
 				name = "Soy Sauce"
-				desc = "A salty soy-based flavoring."
+				desc = "A 
+				y soy-based flavoring."
```
DTraitor: revert this changes

DTraitor (changes requested):
Just one tweak

code/modules/reagents/reagent_containers/food/condiment.dm
githubuser4141: Fixed.

DTraitor (approved):

Vizoplez:
People on their way not to realise 1u of potassium and 1u of water instantly clears your bloodstream and stomach......

shaking my fucking head right now

PR #2199, opened 2022-10-22
tweaks to loadouts of engineers and miners

Fikou:
engineers and miners now use duty boots instead of work boots and sneakers
engineers get thick gloves in their lockers
engineers no longer have ugly blue hardhats

they look really ugly currently and that should make it better


dawsonkeyes:
the advanced rig was intentionally in the locker as people can spawn with rigs like the civie/hacker rig via loadouts, and thus having such a thing would override and delete it

this is now a problem ingame

karlodayo:
> the advanced rig was intentionally in the locker as people can spawn with rigs like the civie/hacker rig via loadouts, and thus having such a thing would override and delete it
> 
> this is now a problem ingame

Agreed, I'd say to please revert this change within your PR.

PR #2200, opened 2022-10-23
Trollge

Vizoplez:
This makes xenoarch FAR more dangerous and helps prevent future atmos grief with it.


Vizoplez:
Essentially this PR
-Buffs the chance of LARGE artifacts spawning
-Makes them more likely to activate on being discovered (thus performing their effect)
-Removes the atmospheric grief effects (such as spawning phoron/plasma)
-Removes the requirement to change the atmosphere to cause them to activate (as this can be construed as potentially breaking server rules, especially on the Ishi where an artifact that suddenly heats the atmosphere to 9000c is bad)

code/modules/xenoarcheaology/effects/radiate.dm
```diff
@@ -5,7 +5,8 @@
 
 /datum/artifact_effect/radiate/New()
 	..()
-	radiation_strength = rand(10, 50)
+	radiation_range = rand(5,15) //This used to NOT have a limit thus would irradiate entire Z-levels
+	radiation_strength = rand(50, 150) //You're GOING to die VERY fast without immediate medical attention
```
DTraitor, line 9: Lets not make radiation too powerful, Lets keep it to something like rand(20, 70)

DTraitor (changes requested):
Just one tweak, otherwise good

Ketrai:
I am still concerned with how frequent artefacts spawn now as that's a lot of free materials for science to abuse. I see a lot more material sheet spawns...

PR #2239, opened 2023-02-01
25 -> 100 miners

chesse20:
Simple web-edit that makes it so theres 100 slots for miners because i heard someone complaining there wasn't enough slots to play as human on discord
🆑 
balance: 100 miners means no possibility of running out of roles because 150 people decided to play the game in one day
/🆑


DTraitor:
Yollo

PR #2257, opened 2023-02-05
Implements click limit per second and per minute. Also fixes a bug that prevents one person in a round from having their preferences loaded

DTraitor:


DTraitor:
Seems to work

PR #2273, opened 2023-02-10
[DNM] Replaced job controller with a proper subsystem. Added playtime tracking

DTraitor:
Replaced job controller with a proper subsystem (`SSjobs`)
Implemented `SSexperience` that tracks players playtime
Added a new table to the DB (`V007`)


sql/migrate/V006__DS13.sql
```diff
@@ -74,6 +74,19 @@ CREATE TABLE IF NOT EXISTS `database`.`store_schematics` (
...
+
+DROP TABLE IF EXISTS `role_time`;
+
+CREATE TABLE `role_time` (
+  `ckey` VARCHAR(32) NOT NULL ,
+  `job` VARCHAR(32) NOT NULL ,
+  `minutes` INT UNSIGNED NOT NULL,
+  PRIMARY KEY (`ckey`, `job`))
+ENGINE = InnoDB;
+
```
francinum: This should really be moved to a new Migrate file, these are intended to be run sequentially and never updated. They're like github commits for the database structure.

francinum (commented):
Please give this a PR body, I know support for this is winding down but things like DB updates are breaking.

PR #2281, opened 2023-02-15
Mute necrochat verb + Timelock

DTraitor:
## Changelog
* Timelocked the Marker role. You need to play 1.5 hour as a signal to control the marker. Timelock doesn't work when there are no players with over 1.5 hour of playtime as signal.
* If Marker player is AFK for more than 5 minutes another can replace him (use 'Become Master Signal' verb).
* Added 'Toggle Reciving Necrochat' verb for admins.


DTraitor:
I hope it fixes an issue...

PR #2291, opened 2023-02-26
Reduced atmo and apc grief and adds Heavy CEC to patron whitelist

blackwolf602:
### Reduced grief and adds Heavy CEC to patron whitelist

- adds Antique Heavy-duty CEC RIG to patron whitelist
- removes Napalm and Phlogiston to prevent chemist from "accidentally" atmos griefing
- changes number of change apc destroy when the engine is destroyed from 5 to 1 (5 destroys a large amount of apc on the map)

it's not giving me the writing guideline on this so looks funky


DTraitor:
Yolo

